PR #19, opened 2020-10-10
[WIP] Run tests with github action

peterrum:


mschreter:
Oh, that's cool :). Thank you.

peterrum:
Test has passed, see https://github.com/mschreter/MeltPoolDG/actions

We should consider to reduce the size of the test (less cells/less time steps): the one test took 575.37 seconds.

PR #28, opened 2020-10-23
Add simplex support

peterrum:
... and many other updates (sorry for putting everything into one PR):
- restructuring of the test folder organization (copy of program is not needed - you can specify which program you want to be tested)
- avoid the usage of `MatrixFree` if possible, since it can be only used for tensor-product elements (and not simplices) -> I'll collect my thoughts about `ScratchData` tomorrow


include/meltpooldg/interface/scratch_data.hpp
```diff
@@ -165,21 +168,28 @@ namespace MeltPoolDG
...
     void
     initialize_dof_vector(VectorType &vec, const unsigned int dof_idx = 0) const
     {
-      matrix_free.initialize_dof_vector(vec, dof_idx);
+      if (do_matrix_free)
+        matrix_free.initialize_dof_vector(vec, dof_idx);
+      else
+        vec.reinit(get_locally_owned_dofs(dof_idx),
+                   get_locally_relevant_dofs(dof_idx),
+                   get_mpi_comm(dof_idx));
```
peterrum, line 192: This is not particular nice now. For simplices, I cannot use `MatrixFree::initialize_dof_vector` so I have to use the alternative version. Which version to use is handled currently by setting the value of `do_matrix_free` in the constructor. I am not happy with this solution :( but it fixes the test...

peterrum:
The failing test is related to https://github.com/dealii/dealii/pull/11027 still has to be merged (it has been approved already twice - so it should be soon).

mschreter:
> The failing test is related to [dealii/dealii#11027](https://github.com/dealii/dealii/pull/11027) still has to be merged (it has been approved already twice - so it should be soon).

Thanks for efforts, Peter. How should we proceed on that? Should we merge it anyway?

include/meltpooldg/interface/scratch_data.hpp
mschreter: I was thinking about whether we should change the boolean from "do_matrix_free" to "do_simplex" but I think "do_matrix_free" makes much more sense. Thus, it is fine for me.

tests/rotating_bubble_mf/rotating_bubble.json
```diff
@@ -0,0 +1,34 @@
+{
```
mschreter, line 1: I really like the new test structure :). Many thanks for that!

peterrum: You're welcome!

peterrum:
I am fine with merging: the PR should be part of dealii/master soon and we know the reason for the failing test. I will extend the other modules to simplex support in follow-up PRs in the next days.

tests/rotating_bubble_mf/rotating_bubble.json
peterrum: Maybe now you understand my comment in https://github.com/MeltPoolDG/MeltPoolDG/pull/22#discussion_r509247769.

peterrum:
I have added to all other modules of MeltPoolDG simplex support so that we can run a full matrix-based level-set simulation with simplex mesh. The result of the rotating bubble looks like this:

rotating-bubble-simplex

Ready to be merged from my side. Thanks for letting me mess around with your code - it feels good that the boring work of a couple moths is actually working in practice! I hope I will have time to extend MatrixFree with simplex support in the next few months.

mschreter:
Many thanks for your great efforts Peter. I am very happy that the extension to simplices worked out and that MeltPoolDG now supports simplex meshes, yey :). Very much appreciated! I am sure that simplex meshes will be a great benefit for the deal.ii community.

peterrum:
I have rerun the tests with the current master of deal.II. All tests are passing now.

PR #33, opened 2020-10-30
Spurious currents

peterrum:


peterrum:
The setup is similiar to 
- https://github.com/MeltPoolDG/adaflo/blob/master/tests/spurious_currents.cc
- https://github.com/MeltPoolDG/adaflo/blob/master/tests/spurious_currents_ls.prm

As far as I can see the setup is fine. Nevertheless, I get some anomalies at the boundaries. @mschreter Maybe you have more success to find the reason.

PR #53, opened 2020-12-15
CI: Build shared library

peterrum:


peterrum:
At least, this seems to fix linking :+1:

PR #58, opened 2020-12-15
Reuse MatrixFree in NavierStokes and enable simplices

peterrum:
This PR allows us to have direct access to `NavierStokesinitialize_matrix_free()`.

### DIFF
--- a/include/meltpooldg/flow/adaflo_wrapper.hpp
+++ b/include/meltpooldg/flow/adaflo_wrapper.hpp
@@ -38,9 +38,7 @@ namespace MeltPoolDG
         : dof_handler_meltpool(scratch_data.get_dof_handler(idx))
         , timer(std::cout, TimerOutput::never, TimerOutput::wall_times)
         , navier_stokes(base_in->parameters.adaflo_params.get_parameters(),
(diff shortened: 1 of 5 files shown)


include/meltpooldg/flow/adaflo_wrapper.hpp
```diff
@@ -70,7 +70,38 @@ namespace MeltPoolDG
...
+
+        scratch_data.attach_constraint_matrix(navier_stokes.get_constraints_u());
+        scratch_data.attach_constraint_matrix(navier_stokes.get_constraints_p());
+
+        const unsigned int quad_index_u = scratch_data.attach_quadrature(
+          QGauss<dim>(base_in->parameters.adaflo_params.get_parameters().velocity_degree + 1));
+        const unsigned int quad_index_p = scratch_data.attach_quadrature(
+          QGauss<dim>(base_in->parameters.adaflo_params.get_parameters().velocity_degree));
+
+        scratch_data.build();
```
peterrum, line 92: Doing it here is not nice. We should discuss how to split up this function. The discussion will be important also in the light of AMR.

peterrum (commented):
@mschreter What do you think!?

PR #60, opened 2020-12-16
"Fix" convergence issue

peterrum:
This is the guilty guy:
https://github.com/MeltPoolDG/MeltPoolDG/blob/b67627f8a620232d1c92d5340aa56dda7bb7dbb9/include/meltpooldg/flow/two_phase_flow_problem.hpp#L190

We should get the constraint matrix from `NavierStokes` as proposed by #58:
https://github.com/MeltPoolDG/MeltPoolDG/blob/b33a2d67b6ef416e83419bcbf352272e925f3337/include/meltpooldg/flow/adaflo_wrapper.hpp#L82

... not yet there ;)

### DIFF
--- a/include/meltpooldg/flow/adaflo_wrapper.hpp
+++ b/include/meltpooldg/flow/adaflo_wrapper.hpp
@@ -79,6 +79,7 @@ namespace MeltPoolDG
       void
       solve() override
       {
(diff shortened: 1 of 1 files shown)


mschreter:
Good news in the morning! Thank you Peter. It remembers me to the bug in the advection diffusion problem. And in the end, it was a one-liner ... I am very glad we made progress on that :1st_place_medal:

PR #77, opened 2021-01-25
Fix amr issues

mschreter:
This PR fixes the amr issues for the level set with evaporation problem. The result looks now the following:

animation


peterrum (approved):
Great! Let's first merge the other PR and than merge this mini PR!

PR #83, opened 2021-02-08
Example evaporating droplet

mschreter:
This PR introduces an additional example file, where a droplet evaporates. For this simple benchmark example, an analytical solution for the reduction of the diameter d exists

<p align="center">
 
</p>

with d, d0, \dot{m}, \rho_v, t denoting the current diameter, the initial diameter, the evaporative mass flux rate, the density of the vapor and the current time, respectively.

### DIFF
--- a/include/meltpooldg/flow/two_phase_flow_problem.hpp
+++ b/include/meltpooldg/flow/two_phase_flow_problem.hpp
@@ -91,6 +91,8 @@ namespace MeltPoolDG::Flow
             {
               level_set_operation.update_normal_vector();
 
+              // @todo: why needed?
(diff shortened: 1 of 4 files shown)


include/meltpooldg/flow/two_phase_flow_problem.hpp
```diff
@@ -91,6 +91,8 @@ namespace MeltPoolDG::Flow
             {
               level_set_operation.update_normal_vector();
 
+              // @todo: why needed?
```
peterrum, line 94: ?

peterrum (approved):

PR #84, opened 2021-02-16
Fix spurious currents

mschreter:
This PR improves the spurious currents simulation by computing the initial smooth level set field from the analytical solution of the reinitialization. An additional parameter `ls n initial reinit steps` is introduced that enables to control the number of initial reinitialization steps.


peterrum (approved):
Thanks!

Issue #81, opened 2021-02-03
Bug in VectorTools


peterrum:
I think the following lines in `VectorTools::convert_fe_sytem_vector_to_block_vector()` and `VectorTools::convert_block_vector_to_fe_sytem_vector()` are wrong (in the general case): 

https://github.com/MeltPoolDG/MeltPoolDG/blob/6621f6083ab444f54916c311cfc2af7411c79bc2/include/meltpooldg/utilities/vector_tools.hpp#L37

https://github.com/MeltPoolDG/MeltPoolDG/blob/6621f6083ab444f54916c311cfc2af7411c79bc2/include/meltpooldg/utilities/vector_tools.hpp#L74

We should use `FiniteElement::component_to_system_index()` for this purpose.

PR #88, opened 2021-02-25
Fix Issue #81

mschreter:
This PR should fix issue #81. At the moment we are not using these utility functions.


peterrum (approved):
Thanks! I totally forgot about this one^^

PR #90, opened 2021-02-26
Evapor additions melt pool

mschreter:
This PR includes:

- fix melt pool with evaporation (no recoil pressure force is added in case of evaporation) 
- introduction of an additional case for computing the density in the interfacial region consistent with the mass balance
- update to C++20 standard

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,7 @@ MESSAGE("_______________________________________________________________________
 
 # this is the standard deal.II search mechanism, including check for Trilinos and p4est
 cmake_minimum_required(VERSION 3.3)  
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 ## avoid in-source compilation
 set(DEFAULT_OUT_OF_SOURCE_FOLDER "build")
--- a/include/meltpooldg/flow/two_phase_flow_problem.hpp
+++ b/include/meltpooldg/flow/two_phase_flow_problem.hpp
@@ -254,23 +254,6 @@ namespace MeltPoolDG::Flow
           "  this->attach_initial_condition(std::make_shared<MyInitializeFunc<dim>>(), 'level_set') "));
 
       /*
-       *    initialize the melt pool operation class
-       */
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ MESSAGE("_______________________________________________________________________
 
 # this is the standard deal.II search mechanism, including check for Trilinos and p4est
 cmake_minimum_required(VERSION 3.3)  
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
peterrum, line 14: What is the reason to go to C++20?

include/meltpooldg/flow/two_phase_flow_problem.hpp
```diff
@@ -518,24 +518,30 @@ namespace MeltPoolDG::Flow
 
               for (unsigned int q = 0; q < ls_values.n_q_points; ++q)
                 {
-                  const auto indicator = parameters.flow.variable_properties_over_interface ?
-                                           ls_values.get_value(q) :
-                                           UtilityFunctions::heaviside(ls_values.get_value(q), 0.5);
+                  const auto indicator =
+                    parameters.flow.variable_properties_over_interface == "true" ||
+                        parameters.flow.variable_properties_over_interface ==
+                          "consistent_with_evaporation" ?
```
peterrum, line 524: Let's do this string compare just once (outside of the loops).

include/meltpooldg/interface/parameters.hpp
```diff
@@ -674,7 +675,7 @@ namespace MeltPoolDG
         prm.add_parameter(
           "laser variable properties over interface",
           laser.variable_properties_over_interface,
-          "Set this parameter to true to interpolate the thermal properties over the interface smoothly.");
+          "Set this parameter to true interpolate the thermal properties over the interface smoothly.");
```
peterrum, line 678: Have you removed the wrong "to"?

peterrum (approved):

CMakeLists.txt
mschreter: I have also activated C++20 standard in deal.II. What are your concerns?

PR #92, opened 2021-03-04
Disable ctest

peterrum:


mschreter:
We should discuss this point in our next meeting. In the meantime I'll merge it since I am also annoyed by the red crosses ^^.

peterrum:
> I am also annoyed by the red crosses ^^.

Same here!

PR #93, opened 2021-03-09
Checkout supported deal.II version in README

nmuch:
When building MeltPoogDG is noticed some problems with the README.MD.

README tells you to install deal.II 9.2 but MeltPoolDG uses features that are only available on the current developer version of deal.II. I changed the README so it checks out a supported commit (today's version) on the deal.II master branch. 

The flag for enabling deal.II simplex support must be set to ON.

Also, the example simulation command had the wrong path to vortex_bubble.json.


README.md
```diff
@@ -11,11 +11,19 @@ This project depends on the following third-party libraries:
 - adaflo (optional)
 
 Before installing deal.II, you have to install the two libraries p4est and Trilinos. More information for that you can find here for p4est (https://www.dealii.org/9.2.0/external-libs/p4est.html) and here for Trilinos (https://www.dealii.org/9.2.0/external-libs/trilinos.html).  
+
+Currently, we use the current developer version of deal.II. Clone the deal.II repository and checkout a working commit:
+
+```bash
+git clone git@github.com:dealii/dealii.git
+cd dealii
+git checkout -b current-4C-version 98eded21ff59f9cb5f1f2ff25e8c4c1b59f7692b
```
mschreter, line 20: MeltPoolDG is (should be) compatible with current deal.II master so I suggest to remove this line. What kind of problems did you observe? Also, the commit hash will be outdated as soon as features from newer deal.II versions are used.
```suggestion
```

mschreter (approved):
Thank you for your changes, it's always good to have new people in :-).

README.md
nmuch: Ok, I'll change it so the current mater branch is checked out. So basically remove this line and the line above. 

I have an older version of the deal.II master branch installed for another project. Using that for MeltPoolDG, some includes of deal.II simplex file did not work; e.g:
```
/include/meltpooldg/advection_diffusion/advection_diffusion_problem.hpp:18:10: fatal error: deal.II/fe/fe_simplex_p.h: No such file or directory
```

I guess some simplex methods have been refactored in the past weeks/months..

peterrum: Yes it has changed (12 days ago): see https://github.com/MeltPoolDG/MeltPoolDG/pull/89. We try to keep adaflo and MeltPoolDG up to data with dealii/master so that we can use the newest bug fixes and features from there (the next one is the marching-cube algorithm: https://github.com/dealii/dealii/pull/11828).

PR #94, opened 2021-03-12
Extensions heat transfer

mschreter:
This PR includes:
- bugfix in smooth interface velocity
- some preparations for the two-phase heat transfer operation
- minor clean-up


include/meltpooldg/evaporation/evaporation_operation.hpp
```diff
@@ -210,7 +211,8 @@ namespace MeltPoolDG::Evaporation
                                      bool               zero_out)
     {
       evaporative_mass_flux.update_ghost_values();
-      normal_vector.update_ghost_values();
+      
+      double mass = 0.0;
 
       scratch_data->get_matrix_free().template cell_loop<VectorType, VectorType>(
         [&](const auto &matrix_free,
```
peterrum, line 218: This is not exactly (thread) safe (the lambda past to `MatrixFree` might be run by multiple threads):
- option 1: make `mass` atomic (see https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add)
- option 2: you perform the loop over the cells yourself
- option 3: disable threads in `MatrixFree::AddtionalData` (in `ScratchData`)

include/meltpooldg/flow/surface_tension_operation.hpp
```diff
@@ -155,18 +155,22 @@ namespace MeltPoolDG::Flow
...
-                             "The surface tension coefficient tends to be negative in "
-                             "some regions. Check the value of the temperature dependent surface "
-                             "tension coefficient."));
+                    {
+                      alpha[v] = alpha[v] <= 0.0 ? 0.0 : alpha[v];
+                      Assert(alpha[v] >= 0.0,
+                             ExcMessage(
+                               "The surface tension coefficient tends to be negative in "
+                               "some regions. Check the value of the temperature dependent surface "
+                               "tension coefficient."));
```
peterrum, line 172: The assert is not need anymore?

peterrum (approved):
Some minor points.

include/meltpooldg/evaporation/evaporation_operation.hpp
mschreter: Would it be thread-safe if I declare `mass` inside the loop?

include/meltpooldg/flow/surface_tension_operation.hpp
mschreter: Nope, actually not.

include/meltpooldg/evaporation/evaporation_operation.hpp
peterrum: Yes and no.Yes it would thread-safe. But the sum would be only complete if all cells are processed in one go, which is not guaranteed (normally it the loop is split up into at lest 3 subregions to enable overlap of communication and computation).

mschreter: Okay, thank you for the explanation. So, since `atomic` is new for me I will read about it first and then include the changes in another PR. Thank you for your comments :-).

peterrum: Atomic means that it is guaranteed either on hardware- or software-level that only a single thread can access (read+write) that variable at a time. If this wouldn't be the case, it might happen that two threads read the same `mass` value before the other has the chance to update it so that one addition is lost. All this is known as "race condition".

peterrum: Actually you have turned threading off:

https://github.com/MeltPoolDG/MeltPoolDG/blob/6d77cf97b9a912011ceab6077e30713964701deb/meltpooldg.cpp#L93

(last number indicates the max number of threads; if you remove that number you will probably get different results)

PR #96, opened 2021-03-16
Add unit test folder

peterrum:


mschreter:
Thank you Peter, looks good to me. 

@nmuch: FYI, since we talked the last time about unit tests. We can use them to test e.g. our utility functions.

mschreter:
One question: They are simply called simultaneously with `ctest`?

peterrum:
> One question: They are simply called simultaneously with ctest?

Yes. If you only want to run these: `ctest -R "unit_tests".`

PR #97, opened 2021-03-23
Work on warnings

peterrum:
... and turn warning to error in CI.


mschreter:
Thank you Peter, good idea.

PR #98, opened 2021-03-23
Remove convert_to_vector()

peterrum:


include/meltpooldg/utilities/vector_tools.hpp
```diff
@@ -1,5 +1,46 @@
...
+  }
+
+  template <typename Number, std::size_t N>
+  dealii::VectorizedArray<Number, N>
+  scalar_product(const dealii::Tensor<1, 1, dealii::VectorizedArray<Number, N>> &vec,
+                 const dealii::VectorizedArray<Number, N> &                      scalar)
+  {
+    return vec[0] * scalar;
+  }
+} // namespace dealii
```
peterrum, line 42: This shows what we would need from deal.II. I am nor sure yet how we will get there^^

mschreter (approved):
Thank you Peter, imho it improves the readability of the code. We might think about adopting it also in adaflo.

PR #100, opened 2021-03-24
[WIP] Add velocity to heat transfer

nmuch:
@mschreter 

This should work now.


mschreter:
This PR introduces the convective term for transporting heat particles with a given velocity field. We added a test file `unidirectional_heat_transfer_with_velocity.json`. In the following, a comparison of the temperature profile for a simulation without velocity and with velocity is shown, where the difference results from the additional transport mechanism:

Screenshot_2021-03-24_13-56-06

PR #101, opened 2021-03-24
Heat transfer: pass velocity with pointer

nmuch:
This PR moves the velocity vector to the heat problem. The heat operation and operator get a pointer to that vector. If the pointer is a nullptr the velocity terms are ignored.

Also, the velocity field function is now handled by the field conditions map.

### DIFF
--- a/include/meltpooldg/heat/heat_transfer_problem.hpp
+++ b/include/meltpooldg/heat/heat_transfer_problem.hpp
@@ -12,33 +12,13 @@ namespace MeltPoolDG::Heat
   using namespace dealii;
 
   template <int dim>
-  class VelocityField : public Function<dim>
-  {
-  public:
-    VelocityField(double velocity)
-      : Function<dim>(dim)
-      , vel(velocity)
-    {}
-
-    double
-    value(const Point<dim> &, const unsigned int component) const override
-    {
-      if (component == 0)
-        return -vel;
-      else
-        return 0.0;
-    }
-
-  private:
-    const double vel;
-  };
-
-  template <int dim>
   class HeatTransferProblem : public ProblemBase<dim>
   {
   private:
     using VectorType      = LinearAlgebra::distributed::Vector<double>;
     using BlockVectorType = LinearAlgebra::distributed::BlockVector<double>;
+
+    VectorType velocity;
 
     TimeIterator<double> time_iterator;
     DoFHandler<dim>      dof_handler;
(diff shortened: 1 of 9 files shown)


mschreter (commented):
In this case, we could remove the parameter "with_velocity" again from `parameters.hpp`.

include/meltpooldg/simulations/unidirectional_heat_transfer/unidirectional_heat_transfer.hpp
```diff
@@ -161,6 +183,10 @@ namespace MeltPoolDG::Simulation::UnidirectionalHeatTransfer
                                        "heat_transfer");
       else
         this->attach_initial_condition(std::make_shared<LinearTemp<dim>>(), "heat_transfer");
+
+      this->attach_velocity_field(std::make_shared<UnidirectionalVelocityField<dim>>(
```
mschreter, line 187: You could also use simply the `dealii::Functions::ConstantFunction` utility

```cpp
std::make_shared<Functions::ConstantFunction<dim>>(this->parameters.heat.velocity, dim)
```

mschreter (approved):

mschreter:
Thank you Nils, looks very good :-), I only have minor comments. Before merging I will decrease the number of the simulated time steps in the new test case, so that the run time of the test is short.

include/meltpooldg/heat/heat_transfer_problem.hpp
```diff
@@ -297,6 +274,24 @@ namespace MeltPoolDG::Heat
...
+    {
+      scratch_data->initialize_dof_vector(velocity, velocity_dof_idx);
+      /*
+       *  set the current time to the advection field function
+       */
+      field_function.set_time(time_iterator.get_current_time());
+      /*
+       *  interpolate the values of the advection velocity
+       */
+
```
mschreter, line 288:
```suggestion
```

include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -85,6 +85,17 @@ namespace MeltPoolDG
...
+    void
+    attach_velocity_field(std::shared_ptr<FunctionType> velocity, const std::string operation_name)
+    {
+      if (!field_conditions_map[operation_name])
+        field_conditions_map[operation_name] = std::make_shared<FieldConditions<dim>>();
+
+      field_conditions_map[operation_name]->velocity_field = velocity;
+    }
+
+
```
mschreter, line 98:
```suggestion
```

include/meltpooldg/simulations/unidirectional_heat_transfer/unidirectional_heat_transfer.hpp
nmuch: Does this not return the same value for every component? Here, we only want the first component to be non-zero.

mschreter: Yes, you're right!

PR #102, opened 2021-03-24
Fix format of Newton output

peterrum:
@mschreter Maybe you could update the test outputs.


mschreter:
Thanks Peter :-). I've updated the test files.

PR #103, opened 2021-03-25
Add two phase flow to heat transfer

nmuch:
This PR introduces a two phase formulation for the heat transfer.

An optional level-set heaviside vector can be added to the heat transfer operator. The material properties are evaluated by the  level-set heaviside value.

The material properties of the second phase are still hard coded in the operator. I'll add two phase material parameter input in a follow-up PR.

I noticed that the nonlinear solver requires more steps to converge if the two phase's material properties differ more. I had to increase the new test case's max. number of iterations.

In the following case the two phases' interface is a horizontal line at y = 0.05.
Screenshot from 2021-03-25 13-32-49


include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -95,6 +95,17 @@ namespace MeltPoolDG
       field_conditions_map[operation_name]->velocity_field = velocity;
     }
 
+    template <typename FunctionType>
+    void
+    attach_level_set_field(std::shared_ptr<FunctionType> level_set_function,
```
peterrum, line 100: I think (initial conditions) of level-set field is handled differently in other parts of the code:

https://github.com/MeltPoolDG/MeltPoolDG/blob/4601e40cd3b9f5d98cf0a2ff9a81c6ba7c74261a/include/meltpooldg/simulations/rotating_bubble/rotating_bubble.hpp#L236

include/meltpooldg/heat/heat_transfer_operation.hpp
```diff
@@ -41,9 +41,14 @@ namespace MeltPoolDG::Heat
     VectorType temperature_old;
     VectorType heat_source;
 
+    // configuration if heat particles are transported with a given velocity field
     const unsigned int vel_dof_idx;
     VectorType *       velocity;
 
+    // configuration if two phases are given
+    const unsigned int ls_dof_idx;
+    VectorType *       level_set_as_heaviside;
```
peterrum, line 50: I am not sure if I understand the comments. What are heat particles? What do you mean by configuration? "Relevant parameters"?

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -423,5 +487,29 @@ namespace MeltPoolDG::Heat
...
+        interp(ls_heaviside_val, /*other cond*/ data.conductivity / 2, data.conductivity);
+      return std::make_tuple(density, capacity, conductivity);
+    }
+
+    template <typename value_type>
+    value_type
+    interp(const value_type &ls, number val1, number val2) const
+    {
+      return (1. - ls) * val1 + ls * val2;
+    }
```
peterrum, line 513: Maybe you could move this function to https://github.com/MeltPoolDG/MeltPoolDG/blob/master/include/meltpooldg/utilities/utilityfunctions.hpp and maybe you could spell out the work `interpolate`.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -313,24 +341,60 @@ namespace MeltPoolDG::Heat
...
+                  const auto ls_heaviside_val = ls_vals.get_value(q_index);
+                  const auto [density, capacity, conductivity] =
+                    get_two_phase_material_parameters(ls_heaviside_val);
+
+                  auto val = density * capacity *
+                               (temp_vals.get_value(q_index) - temp_vals_old.get_value(q_index)) *
+                               this->d_tau_inv -
+                             heat_source_vals.get_value(q_index);
+
+                  val += density * capacity * temp_vals.get_gradient(q_index) *
```
peterrum, line 386: Why not merge these two equations?

peterrum (commented):
Just some minor comments.

include/meltpooldg/heat/heat_transfer_operation.hpp
nmuch: Oh these comments are copied from the heat operator, I didn't really read those.. I think "particles" is a bit misleading. These members are optional in the sense that is the pointers are nullptr these quantities are ignored. I'll improve the comments.

include/meltpooldg/interface/simulationbase.hpp
nmuch: This field function is not an initial condition. In `unidirectional_heat_transfer` the level-set is prescribed by this function at for all time steps because the there is no level-set solver. 

In heat- two-phase flow coupled problems the level-set is determined by the two-phase flow, this function is not required because the level-set is handled by the two-phase flow.

nmuch (commented):
Thank you @peterrum for the review. I'll fix your remarks I didn't comment on.

include/meltpooldg/interface/simulationbase.hpp
mschreter: @peterrum @nmuch I think we will remove this function, as soon as we have really coupled the two-phase flow solver with the heat transfer solver.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -201,25 +205,49 @@ namespace MeltPoolDG::Heat
...
 
-              temp_vals.submit_value(val, q_index);
+                  temp_vals.submit_value(val, q_index);
 
-              auto val_grad = data.conductivity * temp_vals.get_gradient(q_index);
+                  auto val_grad = conductivity * temp_vals.get_gradient(q_index);
 
-              temp_vals.submit_gradient(val_grad, q_index);
+                  temp_vals.submit_gradient(val_grad, q_index);
+                }
```
mschreter, line 250: In this section, there are code duplicates, which would be better to avoid as possible. Maybe you can define the material properties in an if statement as follows and evaluate the operations equally for both cases:
```cpp
VectorizedArrayType conductivity, density, capacity;

if (level_set_as_heaviside)
  [density, capacity, conductivity] =
                    get_two_phase_material_parameters(ls_heaviside_val);
else
{
  density = data.density;  
  capacity = data.capacity;
  conductivity = data.conductivity;
} 
temp_vals.submit_value(....)
temp_vals.submit_gradient(....)
```
In addition, we should put an assert in the constructor to avoid that level set is used without a transport velocity:
```cpp
AssertThrow(level_set_as_heaviside || velocity, ExcMessage("...");
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -313,24 +341,60 @@ namespace MeltPoolDG::Heat
...
+                  temp_vals.submit_value(-val,
+                                         q_index); // negative sign since residual is moved to rhs
 
-              if (velocity)
+                  auto val_grad = data.conductivity * temp_vals.get_gradient(q_index);
+
+                  temp_vals.submit_gradient(-1.0 * val_grad,
+                                            q_index); // -1 since residual is moved to rhs
+                }
+              else // two_phase flow
```
mschreter, line 374: My previous comment also applies to this section.

include/meltpooldg/interface/simulationbase.hpp
peterrum: Since the heat problem is for now only a proxy-problem and the function might be removed later anyway (although we will than also need a similar function in the future so that the test still work), I would simply (mis-)use the other (initial-condition) function. But, I don't have strong feelings about this issue.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -423,5 +487,29 @@ namespace MeltPoolDG::Heat
...
+    get_two_phase_material_parameters(const vaule_type &ls_heaviside_val) const
+    {
+      // todo Input material parameters for two phase
+      const auto density = interp(ls_heaviside_val, /*other dens*/ data.density / 2, data.density);
+      const auto capacity =
+        interp(ls_heaviside_val, /*other cap*/ data.capacity / 2, data.capacity);
+      const auto conductivity =
+        interp(ls_heaviside_val, /*other cond*/ data.conductivity / 2, data.conductivity);
+      return std::make_tuple(density, capacity, conductivity);
+    }
```
mschreter, line 506: Do the hardcoded parameters refer to  `data.capacity/2` etc?

include/meltpooldg/heat/heat_transfer_problem.hpp
```diff
@@ -19,19 +19,23 @@ namespace MeltPoolDG::Heat
     using BlockVectorType = LinearAlgebra::distributed::BlockVector<double>;
 
     VectorType velocity;
+    VectorType level_set_as_heaviside;
```
mschreter, line 22: As mentioned, we should keep in mind to remove this "workaround" again, as soon as we are really solving the two-phase-flow problem but for the moment it is fine and appreciated that you tested it already :-).

mschreter (approved):

mschreter:
Thank you Nils, looks very nice. I have some minor comments. Now we are ready to solve a two-phase-flow-heat-transfer problem, yey :-). I will (hopefully) open a PR on this topic later.

include/meltpooldg/heat/heat_transfer_problem.hpp
```diff
@@ -54,7 +58,14 @@ namespace MeltPoolDG::Heat
             << "t= " << std::setw(10) << std::left << time_iterator.get_current_time();
 
           if (base_in->parameters.heat.velocity != 0.0)
```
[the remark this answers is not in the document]
peterrum: Generally, I would agree in the context of floating-point number. But here, this is a parameter and it will be definelty be zero (if requested in the parameter file) and there are no rounding errors.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: I had some trouble with vectorized array types because the single phase material parameters aren't any. I dicided to make a redundant implementation for the two cases but your solution looks good.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: Yes. This is obviously not a permanent solution.

include/meltpooldg/interface/simulationbase.hpp
mschreter: Yes, you're right. Maybe @nmuch you could remove the `attach_level_set_field` again and use simply `attach_initial_condition` for your testing purpose?

mschreter:
All tests are passing locally, so I will merge this PR.

PR #106, opened 2021-03-26
Heat transfer two-phase: not smearing the material properties

nmuch:
I had a look into the bad convergence of the newton iteration for two-phase heat transfer problems. I varied the velocity and the smeared heaviside interface width.

For the `unidirectional_heat_transfer_with_two_phase` test case I had the following results:

| Velocity  | Interface width | average Newton iterations |
| ------------- | ------------- | ------------- |
| 0.01  | 0.01  | 14 - 15 |
| 0.01  | 0  | 5 |
| 0  | 0.01  | 14 - 15 |
| 0  | 0  | 3 - 4 |

It turns out that making the material properties jump at the interface results in much better convergence.

Currently, this PR introduces an additional parameter for the interface width. However, it might make sense to get rid of interpolated material properties in the heat transfer problem and make the material properties jump at `level_set_heaviside == 0.5`. 

@mschreter wdyt? Should I make the material properties jump at the interface, no matter the interface width?


mschreter:
Thank you Nils, looks interesting. I suggest that you introduce an additional parameter `bool variable_properties_over_interface`, which is default set to false and enables the user to choose. This is similar to what we do in the `two_phase_flow_problem` (although here we distinguish between three options):

https://github.com/MeltPoolDG/MeltPoolDG/blob/21e23661e197d66e7bf69f9e286a74a96c0d9809/include/meltpooldg/flow/two_phase_flow_problem.hpp#L585-L604

include/meltpooldg/interface/parameters.hpp
```diff
@@ -778,6 +779,9 @@ namespace MeltPoolDG
         prm.add_parameter("heat two phase",
                           heat.two_phase,
                           "Set this parameter for two phase flow.");
+        prm.add_parameter("heat two phase interface width",
```
mschreter, line 782: This parameter can be removed again (which we won't need anymore soon anyhow), when you introduce instead the boolean `variable_properties_over_interface`.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -468,18 +468,24 @@ namespace MeltPoolDG::Heat
...
+    get_two_phase_material_parameters(const VectorizedArrayType &ls_heaviside_val) const
     {
       // todo Input material parameters for two phase
-      const auto density      = UtilityFunctions::interpolate(ls_heaviside_val,
+
+      VectorizedArrayType weight;
+      if (data.variable_properties_over_interface)
+        weight = ls_heaviside_val;
+      else
+        weight = UtilityFunctions::heaviside(ls_heaviside_val, 0.5);
```
nmuch, line 480: I introduced this if statement for optional variable parameters over interface. By default, the parameters jump at `level_set_heaviside = 0.5`.

tests/unidirectional_heat_transfer_with_two_phase.mpirun=4.output
```diff
@@ -2,49 +2,49 @@ t= 0.001
...
-             72    6.40585e+02 ✗     2.41723e+00 ✗
-             71    5.12718e-01 ✗     1.83945e-03 ✗
-             63    2.31984e-03 ✗     5.92817e-06 ✗
-             66    5.87598e-06 ✗     1.24965e-08 ✗
-             75    1.96324e-08 ✓     5.08723e-11 ✓
+             73    6.40585e+02 ✗     2.41042e+00 ✗
+             73    3.69236e-01 ✗     1.05115e-03 ✗
+             65    1.14671e-03 ✗     2.18439e-06 ✗
+             69    1.95164e-06 ✗     3.03505e-09 ✗
+             75    6.65212e-09 ✓     1.26347e-11 ✓
```
mschreter, line 9: Why did the values change?

nmuch: I'm not sure. Maybe the implementation in the first commit averaged the phases' properties if the level_set_heaviside value was exactly 0.5. 
The resulting temperature does not differ really.

Do we want to have averaged material properties at `level_set_heaviside == 0.5`?

mschreter: Ah, I just remembered -- we had the discussion about the tests not running on your hardware. I will run the test on my machine and let you know. 

What do you mean with averaged material properties? Something like (phi_l+phi_g)/2? I would say no.

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Looks good, thank you :-).

tests/unidirectional_heat_transfer_with_two_phase.mpirun=4.output
mschreter: > Ah, I just remembered -- we had the discussion about the tests not running on your hardware. I will run the test on my machine and let you know.
> 
> What do you mean with averaged material properties? Something like (phi_l+phi_g)/2? I would say no.

Or can you think of another reason?

mschreter:
All tests are passing :-).

tests/unidirectional_heat_transfer_with_two_phase.mpirun=4.output
nmuch: > What do you mean with averaged material properties? Something like (phi_l+phi_g)/2? I would say no.
Yes, I meant that. The current implementation does not have that average step.

> Or can you think of another reason?
Not really :man_shrugging: the result however did not change notably.

PR #107, opened 2021-03-26
Add two-phase material parameter input for heat transfer

nmuch:
This PR introduces a two-phase material parameter input section for heat transfer. This material parameter section can be expanded for more material parameters in other physical contexts.

@mschreter what do you think of this design? If we always use the material section to define materials we can remove the material_data pointer and pass this struct by reference instead.

This might create a merge conflict with #106. I'll rebase whatever is merged last.

### DIFF
--- a/include/meltpooldg/interface/parameters.hpp
+++ b/include/meltpooldg/interface/parameters.hpp
@@ -239,6 +239,24 @@ namespace MeltPoolDG
   };
 
   template <typename number = double>
+  struct MaterialData
+  {
+    struct Liquid
+    {
+      number capacity     = 0.0;
+      number conductivity = 0.0;
+      number density      = 0.0;
+    } liquid;
+
+    struct Gas
+    {
+      number capacity     = 0.0;
+      number conductivity = 0.0;
+      number density      = 0.0;
+    } gas;
+  };
+
+  template <typename number = double>
   struct ParaviewData
   {
     bool        do_output           = false;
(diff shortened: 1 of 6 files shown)


include/meltpooldg/simulations/unidirectional_heat_transfer/unidirectional_heat_transfer_with_two_phase.json
```diff
@@ -25,9 +22,16 @@
...
-    "paraview print boundary id": "true",
-    "paraview filename": "solution_heat_transfer_two_phase"
+  "material": {
+    "material liquid capacity": "460",
+    "material liquid conductivity": "55.563",
+    "material liquid density": "7850",
+    "material gas capacity": "230",
+    "material gas conductivity": "27.7815",
+    "material gas density": "3925",
+    "paraview": {
```
mschreter, line 32: I think there is a bracelet missing?

include/meltpooldg/interface/parameters.hpp
```diff
@@ -237,6 +237,24 @@ namespace MeltPoolDG
     number ls_value_gas                       = -1.0;
   };
 
+  template <typename number = double>
+  struct MaterialData
+  {
+    struct Liquid
+    {
+      number capacity;
```
mschreter, line 245: Would you mind to provide default values to the parameters?

mschreter:
Thank you Nils, its a good start for the moment. Just to briefly rephrase what we've discussed, in the future we might think about having a `material` object, holding parameters of only one material or also many more phases/materials, which can be accessed by `get_phase_properties("phase_1")` or similar. I will have a look at the tests also from the other PR and will then merge one PR by another.

mschreter:
@nmuch: I simply used the online tool for fixing the merge conflicts. Maybe you can address the minor comments I had and in the meantime I will again have a look at the tests. Thanks a lot, I think we made nice progress this week :+1:.

include/meltpooldg/interface/parameters.hpp
nmuch: Sure

PR #111, opened 2021-03-29
Automatically create paraview output directory

mschreter:
@nmuch: FYI
... already a long time overdue :-).

### DIFF
--- a/include/meltpooldg/utilities/postprocessor.hpp
+++ b/include/meltpooldg/utilities/postprocessor.hpp
@@ -11,6 +11,8 @@
 #include <deal.II/base/conditional_ostream.h>
 #include <deal.II/base/table_handler.h>
 
+#include <filesystem>
+
 using namespace dealii;
 
 // @ todo: !!! clean-up and refactoring !!!
@@ -46,7 +48,19 @@ namespace MeltPoolDG
       , triangulation(triangulation_in)
       , pcout(std::cout, (Utilities::MPI::this_mpi_process(mpi_communicator) == 0))
       , do_simplex(!triangulation.all_reference_cells_are_hyper_cube())
-    {}
+    {
+      // check if the requested paraview directory exists and if not create the directory
+      AssertThrow(!std::filesystem::exists(pv_data.directory) ||
+                    std::filesystem::is_directory(pv_data.directory),
+                  ExcMessage("You are trying to create a folder with the name <" +
+                             std::string(pv_data.directory) +
+                             ">. However, a file with the same name already exists! "
+                             "Possible solutions could be to either rename the output "
+                             "folder or to rename/move the existing file."));
+
+      if (!std::filesystem::exists(pv_data.directory))
+        std::filesystem::create_directory(pv_data.directory);
+    }
 
     /*
      *  This function collects and performs all relevant postprocessing steps.
@@ -82,7 +96,7 @@ namespace MeltPoolDG
       data_out.set_flags(flags);
 
       data_out.build_patches(mapping);
-      data_out.write_vtu_with_pvtu_record(pv_data.directory,
+      data_out.write_vtu_with_pvtu_record(pv_data.directory + "/",
                                           pv_data.filename,
                                           time_step,
                                           mpi_communicator,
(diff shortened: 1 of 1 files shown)


include/meltpooldg/utilities/postprocessor.hpp
```diff
@@ -83,7 +90,7 @@ namespace MeltPoolDG
 
       data_out.build_patches(mapping);
       data_out.write_vtu_with_pvtu_record(pv_data.directory,
-                                          pv_data.filename,
+                                          "/" + pv_data.filename,
```
peterrum, line 93: I don't get this change!

As a reference: https://github.com/kronbichler/adaflo/blob/b1b9e319784573f911b8c3306dfa5d2fc5617610/source/flow_base_algorithm.cc#L241-L248

include/meltpooldg/utilities/postprocessor.hpp
```diff
@@ -46,7 +48,12 @@ namespace MeltPoolDG
       , triangulation(triangulation_in)
       , pcout(std::cout, (Utilities::MPI::this_mpi_process(mpi_communicator) == 0))
       , do_simplex(!triangulation.all_reference_cells_are_hyper_cube())
-    {}
+    {
+      // check if the requested paraview directory exists and if not create the directory
+      if (!std::filesystem::is_directory(pv_data.directory) ||
+          !std::filesystem::exists(pv_data.directory))
```
peterrum, line 54:
```
      if (!std::filesystem::is_directory(pv_data.directory) &&
          !std::filesystem::exists(pv_data.directory))
```
You might have got a file with the same name!?

peterrum (approved):
I will not understand what took C++ so long to add `std::filesystem` (C++17).

include/meltpooldg/utilities/postprocessor.hpp
mschreter: It won't matter, where the slash is located. I can also add it as a suffix to the `pv_data.directory`.

peterrum: Ah. Yes, let's move it to the folder name. The problem is that the called function does not normalize the names (for that one would need `std::filesystem`).

include/meltpooldg/utilities/postprocessor.hpp
mschreter: I removed the second statement.

include/meltpooldg/utilities/postprocessor.hpp
```diff
@@ -46,7 +48,11 @@ namespace MeltPoolDG
       , triangulation(triangulation_in)
       , pcout(std::cout, (Utilities::MPI::this_mpi_process(mpi_communicator) == 0))
       , do_simplex(!triangulation.all_reference_cells_are_hyper_cube())
-    {}
+    {
+      // check if the requested paraview directory exists and if not create the directory
+      if (!std::filesystem::is_directory(pv_data.directory))
+        std::filesystem::create_directory(pv_data.directory);
```
peterrum, line 54: Could you create a file with the name `pv_data.directory` and have a look at what happens?

include/meltpooldg/utilities/postprocessor.hpp
mschreter: Mmh...

```
----------------------------------------------------
Exception on processing: 
filesystem error: cannot create directory: File exists [my_directory]
Aborting!
```

mschreter: ... only if I created a file `my_directory` with `touch my_directory`. If I have a file `my_directory.txt`, it still creates the directory.

include/meltpooldg/utilities/postprocessor.hpp
mschreter: Please find the comments above.

include/meltpooldg/utilities/postprocessor.hpp
```diff
@@ -46,7 +48,19 @@ namespace MeltPoolDG
...
-    {}
+    {
+      // check if the requested paraview directory exists and if not create the directory
+      AssertThrow(!std::filesystem::exists(pv_data.directory) ||
+                    std::filesystem::is_directory(pv_data.directory),
+                  ExcMessage("You are trying to create a folder with the name <" +
+                             std::string(pv_data.directory) +
+                             ">. However, a file with the same name already exists! "
+                             "Possible solutions could be to either rename the output "
+                             "folder or to rename/move the existing file."));
```
peterrum, line 59: A minor suggestion:
```suggestion
                             "folder in the parameter file or to rename/move the existing file."));
```

peterrum (approved):
Thanks!

Issue #108, opened 2021-03-26
Distance functions (initial level set) for mimicking powder particles


mschreter:
It would be nice to introduce two additional features in view of modeling the powder particles:

- [ ] Signed distance function that creates certain regular distribution of powder particles (characterized by phi=1):
![image](https://user-images.githubusercontent.com/70260016/112658920-2d19e180-8e54-11eb-8a9e-50b813402317.png)

- [ ] Enable to "merge" two signed distance functions with priorization:
![image](https://user-images.githubusercontent.com/70260016/112659334-8da91e80-8e54-11eb-9030-2bd3562e3e69.png)

PR #113, opened 2021-03-30
Utility function for merging two level set fields

mschreter:
This PR introduces the utility function to merge two level set fields based on boolean algebra (and solves Issue #108). The following cases are supported:

- intersection
- difference 
- union

image

### DIFF
--- a/include/meltpooldg/utilities/utilityfunctions.hpp
+++ b/include/meltpooldg/utilities/utilityfunctions.hpp
@@ -5,14 +5,26 @@
 #include <deal.II/base/vectorization.h>
 
 #include <deal.II/fe/fe_dgq.h>
+#include <deal.II/fe/fe_q.h>
 #include <deal.II/fe/fe_tools.h>
 
 #include <deal.II/lac/generic_linear_algebra.h>
+
+#include <deal.II/matrix_free/matrix_free.h>
+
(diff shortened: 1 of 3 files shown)


include/meltpooldg/utilities/utilityfunctions.hpp
```diff
@@ -5,15 +5,27 @@
...
 
 #include <deal.II/fe/fe_dgq.h>
+#include <deal.II/fe/fe_q.h>
 #include <deal.II/fe/fe_tools.h>
 
 #include <deal.II/lac/generic_linear_algebra.h>
 
+#include <deal.II/matrix_free/matrix_free.h>
+
+#include <meltpooldg/utilities/fe_integrator.hpp>
```
peterrum, line 15: Not needed.

peterrum (approved):
Very nice!

mschreter:
@peterrum: This PR is ready from my side. Maybe you could have a quick look at it? The necessary clean-up works will be part of a follow-up PR.

include/meltpooldg/utilities/utilityfunctions.hpp
peterrum: I don't see it where it is used. But nevertheless merge ;)

peterrum (approved):

PR #114, opened 2021-03-30
Add thermo capillary two droplet simulation

nmuch:
This PR adds the two-droplet simulation case from [Nas, Tryggvason 2003](https://www.sciencedirect.com/science/article/pii/S0301932203000843?casa_token=Sw9xFN3LLFIAAAAA:6U4NSOa_hNtNYYvYM8d4ctMo_M6XfzqLOYRFTfctMUa5cmOLFKXPisJpgGqvB2fu66AnxL3yWyE). 
Also, the level-set field is now passed to the heat operation.

Here, the result is plotted for t_r = 3.8. The lines are iso-thermals.
image

Some issue still remain:
- The side boundary conditions should be periodic. I'll add periodic boundary conditions in a follow-up PR.
- The top/bottom boundary conditions don't seem to correlate to what's described in the paper.
- The (heat and fluid) material parameters are distributed across multiple sections in the Parameters struct. I'll unify the definition of material parameters for multiple physics using the material struct from #107 in a follow-up PR.
- The simulation result does not correlate to the reference solution after some time. The image below shows the result for  t_r = 25.3. This problem may be resolved by the correct boundary conditions.
image

PR #117, opened 2021-03-31
Add periodic boundary condition

nmuch:
This PR adds a periodic boundary condition `SimulationBase::attach_periodic_boundary_condition()`.

The periodic boundary condition is applied to the thermo capillary two droplet simulation, introduced in #114. The left and right boundaries are periodic. 
In the picture below I display a copy of the result attached to the periodic boundary (indicated by the black line) to demonstrate the periodic boundary. The white lines are streamlines.
grafik


mschreter:
> This PR adds a periodic boundary condition `SimulationBase::attach_periodic_boundary_condition()`.
> 
> The periodic boundary condition is applied to the thermo capillary two droplet simulation, introduced in #114. The left and right boundaries are periodic.
> In the picture below I display a copy of the result attached to the periodic boundary (indicated by the black line) to demonstrate the periodic boundary. The white lines are streamlines.
> grafik

Very nice Nils, thank you. There is a slight offset between the streamlines, but this is probably an artefact. Is this correct?

include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -205,6 +205,30 @@ namespace MeltPoolDG
...
+    attach_periodic_boundary_condition(const unsigned           direction,
+                                       const types::boundary_id id_in,
+                                       const types::boundary_id id_out,
+                                       const std::string        operation_name)
+    {
+      if (!boundary_conditions_map[operation_name])
+        boundary_conditions_map[operation_name] = std::make_shared<BoundaryConditions<dim>>();
+
+      auto &bc = boundary_conditions_map[operation_name]->periodic_bc;
+      bc.push_back(std::make_tuple(direction, id_in, id_out));
```
mschreter, line 224: Maybe we can add an assert here that the direction must be always smaller than dim.

include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -205,6 +205,30 @@ namespace MeltPoolDG
...
+
+      std::vector<GridTools::PeriodicFacePair<typename Triangulation<dim>::cell_iterator>>
+        periodic_faces;
+      GridTools::collect_periodic_faces(*triangulation, id_in, id_out, direction, periodic_faces);
+      triangulation->add_periodicity(periodic_faces);
```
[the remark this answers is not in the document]
mschreter: Probably, they have to enter somehow the `AffineConstraints`.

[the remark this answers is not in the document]
mschreter: I suppose, the following function should help: 

https://github.com/s1lin/deal.ii-8.4.1-Tpetra/blob/36faed9559ec1f80e86369d91e43e7b09d1caa39/include/deal.II/dofs/dof_tools.h#L1087-L1095

peterrum:
A general remark. The line

https://github.com/MeltPoolDG/MeltPoolDG/blob/4f3e7d974dc148eeed877e27cf312d0b424604f2/include/meltpooldg/interface/simulationbase.hpp#L228-L229

indicates that if we add PBC to any physical field it will automatically affect all other fields, since the common triangulation is modified. I guess it make sense that if one field has PBC all fields have it as well? Or does one want to apply PBC for fluid and DBC for Neumann?

> Do we have to handle periodicity somehow in our operations?

If `GridTools::make_periodicity_constraints()` and `DoFTools::make_periodicity_constraints` is used, the operators should work without changes.

nmuch:
> There is a slight offset between the streamlines, but this is probably an artefact. Is this correct?

This looks like an offset because the streamlines aren't drawn across the periodic interface, so the lines don't align perfectly. The actual dof values perfectly match on both sides.

mschreter:
> > Do we have to handle periodicity somehow in our operations?
> 
> No, we don't yet. The example above worked, because adaflo handled this for the flow. I'll add this for our operators, too.
> 
> > I guess it make sense that if one field has PBC all fields have it as well? Or does one want to apply PBC for fluid and DBC for Neumann?
> 
> I think if we have a periodic boundary condition it should apply for all physical field. I can't think of a case where a partially periodic boundary makes physical sense. Wdyt @mschreter?

Yes, we definitely have to add them to all fields (similar to dirichlet constraints). However, there is no need to call

```cpp
    void
    attach_periodic_boundary_condition(const unsigned           direction,
                                       const types::boundary_id id_in,
                                       const types::boundary_id id_out,
                                       const std::string        operation_name)
```

for each operation separately (the triangulation is the same for each operation). So, you could skip the function argument `operation_name` here and simply replace it by `periodic` as a placeholder. Then, each operation can simply call `get_periodic_boundary_condition()` and distribute the constraints accordingly, using `DoFTools::make_periodicity_constraints()`.

If you could not follow my explanations, we can also discuss this in person :-).

nmuch:
I added periodic constraints to all field in the two phase flow problem. I'll add this section to the other problems later.

include/meltpooldg/flow/two_phase_flow_problem.hpp
```diff
@@ -565,15 +556,45 @@ namespace MeltPoolDG::Flow
...
+        ls_hanging_node_constraints,
+        AffineConstraints<double>::MergeConflictBehavior::right_object_wins);
+      temp_constraints_dirichlet.close();
+      temp_constraints_dirichlet.merge(
+        ls_hanging_node_constraints,
+        AffineConstraints<double>::MergeConflictBehavior::right_object_wins);
+      reinit_constraints_dirichlet.close();
+      reinit_constraints_dirichlet.merge(
+        ls_hanging_node_constraints,
+        AffineConstraints<double>::MergeConflictBehavior::right_object_wins);
```
mschreter, line 597: Just as a reminder for myself: The lines for setting up constraints are repeated in every problem. Maybe we can implement an `attach_constraints(AffineConstraints<double>&)` function inside the relevant operations

mschreter:
Thank you @nmuch, it looks very good :+1:. I added one minor comment, which probably refers to a copy-paste error. Could you fix this? After that, I am happy to merge :-).

include/meltpooldg/flow/two_phase_flow_problem.hpp
```diff
@@ -565,15 +556,45 @@ namespace MeltPoolDG::Flow
...
-        ls_hanging_node_constraints,
-        AffineConstraints<double>::MergeConflictBehavior::right_object_wins);
 
       evapor_hanging_node_constraints.clear();
       evapor_hanging_node_constraints.reinit(
         scratch_data->get_locally_relevant_dofs(evapor_vel_dof_idx));
       DoFTools::make_hanging_node_constraints(dof_handler_evapor, evapor_hanging_node_constraints);
+
+      // periodic constraints
+      if (base_in->get_bc("periodic") && !base_in->get_dirichlet_bc("periodic").empty())
```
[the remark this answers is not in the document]
nmuch: Oh yes, this is a copy oversight 😅

nmuch:
I fixed the bug with setting periodic constraints in the two-phase flow problem. The periodicity should be applied to alle fields now.

However, I get a error message when running the two-phase periodic problem with 4 or more mpi processes:
> [] *** An error occurred in MPI_Waitall
[] *** reported by process [2149777409,1]
[] *** on communicator MPI_COMM_WORLD
[] *** MPI_ERR_TRUNCATE: message truncated
[] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[] ***    and potentially your MPI job)

The error occurs in `HeatTransferOperator::tangent_boundary_loop()`. @peterrum @mschreter do you have any idea what's wrong here? I think it's weird it works with 1 to 3 mpi procs.

peterrum:
@nmuch How can we replicate your problem. Is there a particular json file we should try out?

nmuch:
> How can we replicate your problem. 

I ran meltpooldg with mpirun and `include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.json` and got the error.

peterrum:
> I ran meltpooldg with mpirun and include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.json and got the error.

I don't get any issues. Maybe we can zoom later so that we can isolate the problem.

include/meltpooldg/flow/adaflo_wrapper.hpp
```diff
@@ -45,6 +45,11 @@ namespace MeltPoolDG::Flow
         {
           for (const auto &symmetry_id : base_in->get_symmetry_id("navier_stokes_u"))
             navier_stokes.set_symmetry_boundary(symmetry_id);
+          for (const auto &periodic_bc : base_in->get_periodic_bc())
+            {
+              const auto [id_in, id_out, direction] = periodic_bc;
+              navier_stokes.set_periodic_direction(direction, id_in, id_out);
+            }
```
mschreter, line 52: @nmuch: Regarding your problem, I am getting currently a segmentation fault if I run e.g. `mpirun -np 4 meltpooldg ../tests/recoil_pressure.json`. When I comment out the 5 lines above this comment then everything is fine. The behavior of `base_in->get_periodic_bc()` might be undefined when the map is empty. Maybe you can reproduce this behavior.

nmuch: Hmm I can reproduce that behavior. I'll have a look into this issue.

nmuch: These boundary condition getter are usually guarded by if conditions whether the boundary condition is empty or not. Here, the `get_periodic_bc()` call was not guarded correctly. It should be fixed now.

mschreter:
@nmuch: I've adapted the concept for PBC, by storing them in an individual object, independent from the `operation_name`. I hope this is fine for you as well.

I've noticed that the function call
```cpp
GridTools::collect_periodic_faces(*triangulation, id_in, id_out, direction, periodic_faces);
```
takes extremely long. @nmuch: Did you also observe the same? @peterrum: Do you have any explanations for that? The behavior can be reproduced by calling 
```
mpirun -np 30 meltpooldg ../include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.json 
```

peterrum:
> I've noticed that the function call GridTools::collect_periodic_faces(*triangulation, id_in, id_out, direction, periodic_faces);
takes extremely long. @nmuch: Did you also observe the same? @peterrum: Do you have any explanations for that? The behavior can be reproduced by calling

My guess is: https://github.com/dealii/dealii/blob/ccc2ebc0754ce6922c0f41135bf9a1f7d1d4d256/source/grid/grid_tools_dof_handlers.cc#L2058

I think you could make this faster by applying PBC on the coarse level and then refine the mesh (but not sure).

mschreter:
> > I've noticed that the function call GridTools::collect_periodic_faces(*triangulation, id_in, id_out, direction, periodic_faces);
> > takes extremely long. @nmuch: Did you also observe the same? @peterrum: Do you have any explanations for that? The behavior can be reproduced by calling
> 
> My guess is: https://github.com/dealii/dealii/blob/ccc2ebc0754ce6922c0f41135bf9a1f7d1d4d256/source/grid/grid_tools_dof_handlers.cc#L2058
> 
> I think you could make this faster by applying PBC on the coarse level and then refine the mesh (but not sure).

Thank you, I could fix this issue by moving `triangulation.refine()` after `collect_periodic_faces

mschreter (approved):

include/meltpooldg/flow/two_phase_flow_problem.hpp
```diff
@@ -361,15 +361,18 @@ namespace MeltPoolDG::Flow
       /*
        *    initialize the heat operation class
        */
-      // @todo: fill level set
       if (base_in->parameters.base.problem_name == "two_phase_flow_with_heat_transfer")
         {
+          /**
+           * @todo[PM]: print to level 0 -> needs test output to be updated
+           */
+          scratch_data->get_pcout(1) << "setup heat problem" << std::endl;
```
peterrum, line 369: I have move this into level 1 output so that the tests pass without needing to adjust them. But I guess level 0 would be better.

include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -346,5 +378,6 @@ namespace MeltPoolDG
   private:
     std::map<std::string, std::shared_ptr<FieldConditions<dim>>>    field_conditions_map;
     std::map<std::string, std::shared_ptr<BoundaryConditions<dim>>> boundary_conditions_map;
+    PeriodicBoundaryConditions<dim>                                 periodic_boundary_conditions;
```
peterrum, line 381: This used to be a shared pointer. But then we would get a segfault during `get_periodic_bc()` if `attach_periodic_boundary_condition()` was not called.

peterrum (commented):
Looks good to me. I have directly added some minor modifications to this branch.

@nmuch Could you check if you still get the MPI problem. Thx. Once that issue is cleared, this PR should be fine to merge.

include/meltpooldg/flow/two_phase_flow_problem.hpp
mschreter: Sorry, this was a leftover from my side from debugging. I'll remove this.

mschreter:
> Looks good to me. I have directly added some minor modifications to this branch.
> 
> @nmuch Could you check if you still get the MPI problem. Thx. Once that issue is cleared, this PR should be fine to merge.

When I introduced the `PeriodicBoundaryCondtions` class, I played a little bit around with the `thermo_capillary_two_droplets` example, and I observed segfaults. I had the following lines under suspision:

https://github.com/MeltPoolDG/MeltPoolDG/blob/16b496275902bf53c9a10b176a24d1db31c39b8c/include/meltpooldg/reinitialization/reinitialization_operation_adaflo_wrapper.hpp#L81-L95

By including https://github.com/MeltPoolDG/MeltPoolDG/commit/8ef5d9a1c46925088d0aa5e9c1aa51fd60a33a37#diff-855a227a0ef71f10d88df399be8520556d1d12818e1a4ce9412491a015403993, these issues were solved. Maybe that could be related to your problems, @nmuch.

peterrum:
This looks more like an issue we have introduced in the previous PR, but maybe I am wrong...

nmuch:
Thank you @mschreter @peterrum for looking into this PR. Sadly I still have some issues:
The MPI problem https://github.com/MeltPoolDG/MeltPoolDG/pull/117#issuecomment-811762113 seems to be fixed for debug, but it still remains in release.
Also, I noticed, that the mesh refinement is not applied to the two-droplet case correctly. I'll have a look into fixing that.

peterrum:
I don't get any issues in release mode. But one guess would be that the following lines:

https://github.com/MeltPoolDG/MeltPoolDG/blob/519511114be3f4b88bebf2ec817465bab901c6c9/include/meltpooldg/heat/heat_transfer_operator.hpp#L166-L169

make problems. Is it possible that one of these if-statements (or similar ones) returns true on one process and false on an other one (due to some undefined behavior)? That you see the problem in `HeatTransferOperator::tangent_boundary_loop()` is just a coincidence and is probably raised in https://github.com/dealii/dealii/blob/2a8b880ac45aed20c14c4b36da8274a33d33a4cf/include/deal.II/base/partitioner.templates.h#L197-L199 during the matrix-free loop, since a communication step has not been finalized before.

nmuch:
Thanks to @peterrum the simulation now works perfectly on my workstation.

mschreter:
Thank you @nmuch and @peterrum, these are great news! The failing CI is related to PR https://github.com/kronbichler/adaflo/pull/42.

peterrum:
I have restarted CI. @mschreter so you want to merge once it is happy.

PR #115, opened 2021-03-30
Fix test infrastructure

mschreter:
@peterrum: FYI, preparations for tomorrow

Closes #116

PR #118, opened 2021-04-02
Clean-up from fixing test infrastructure

mschreter:
This PR solves the issues suggested in the review of PR #115.


include/meltpooldg/interface/scratch_data.hpp
```diff
@@ -46,15 +46,17 @@ namespace MeltPoolDG
     using BlockVectorType = LinearAlgebra::distributed::BlockVector<number>;
 
   public:
-    ScratchData(const unsigned int max_verbosity_level = 0, const bool do_matrix_free = true)
+    ScratchData(const unsigned int max_verbosity_level = 0,
+                const bool         do_matrix_free      = true,
+                const MPI_Comm     mpi_communicator    = MPI_COMM_WORLD)
```
peterrum, line 51: I would prefer if we would be explicit regarding the communicator (this is normally our policy in deal.II). I think we have access to the communicator where we create `ScratchData`, don't we?

peterrum (approved):
Thanks a lot! My comments did not imply that you should do this ;)

mschreter:
> Thanks a lot! My comments did not imply that you should do this ;)

I am so happy right now, so I decided to fix this immediately :)

include/meltpooldg/interface/scratch_data.hpp
mschreter: Okay, maybe we can discuss this another time. I''ve opened an issue relating to this topic https://github.com/MeltPoolDG/MeltPoolDG/issues/119.

PR #125, opened 2021-04-08
Two-droplet simulation case with new parameters

nmuch:
I had another look into the parameters of the two thermo-capillary droplets simulation case by [Nas and Tryggvason](https://doi.org/10.1016/S0301-9322(03)00084-3). I changed the case to the "Light drops" case, where the factor between material properties is 0.5. This case has been validated by [Balcázar et al.](https://doi.org/10.1016/j.ijheatfluidflow.2016.09.015) and they also provide an estimate for the time step size.
With these changes the droplets deform very little like observed by Nas and Balcázar. In contrast, the droplets where very unstable with the parameters in #117.

However, the solution of the simulation changes drastically from the literature's solution from the moment the droplets touch. We allow the droplets to merge, whereas Nas and Balcázar don't.

solution at 20t* - lines are isothermals
image

solution at 40t* - lines are isothermals
image


tests/thermo_capillary_two_droplets.json
```diff
@@ -0,0 +1,89 @@
...
+          "lin velocity preconditioner": "ilu scalar",
+          "lin pressure mass preconditioner": "ilu",
+          "lin its before inner solvers": "50"
+        }
+      },
+      "Time stepping": {
+        "scheme": "bdf_2"
+      },
+      "Output options": {
+        "output verbosity": "2",
```
peterrum, line 58:
```suggestion
        "output verbosity": "2",
```

mschreter:
@nmuch: Looks like we are approaching :-), nice! Could you tell me from which publication you took the two figures above? Unfortunately, I could not find them in the mentioned one by Balcázar et al. Thanks!

nmuch:
>  Could you tell me from which publication you took the two figures above?

Those are the figures by  [Nas and Tryggvason](https://doi.org/10.1016/S0301-9322(03)00084-3) (Fig. 8)

mschreter:
> > Could you tell me from which publication you took the two figures above?
> 
> Those are the figures by [Nas and Tryggvason](https://doi.org/10.1016/S0301-9322(03)00084-3) (Fig. 8)

Regarding the coalescence of the two bubbles: Nas and Tryggvason used front-tracking for the two-phase flow (the interface is represented as a surface grid, which is simply advected by the flow velocity), where imho special treatment is needed to model topology changes such as the coalescence of particles (see e.g.[Razizadeh et al. 2018]( https://link.springer.com/content/pdf/10.1007/s00707-017-1958-5.pdf). While Balcázar et al. used the level set method, however, they introduced an additional level set for the second bubble (strictly speaking three-phase flow) to avoid coalescence on purpose. Hence, it is not surprising for me that the bubbles coalesce in our case.

include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.hpp
```diff
@@ -51,39 +55,46 @@
...
   using namespace dealii;
   using namespace MeltPoolDG::Simulation;
 
-  static constexpr double radius                = 0.043817805;
+  static constexpr double radius                = 0.048;
   static constexpr double x_outer               = 8. * radius;
   static constexpr double z_outer               = 16. * radius;
-  static constexpr double reference_temperautre = 290.;
-  static constexpr double temperature_gradient  = 4.754459964;
+  static constexpr double reference_temperautre = 0.0;
```
mschreter, line 96: typo --> temperature

include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.hpp
```diff
@@ -17,7 +17,11 @@
  *
  * Nas, S., & Tryggvason, G. (2003). Thermocapillary interaction of two bubbles or drops.
```
mschreter, line 18: Could you add a `fix_pressure_constant` boundary condition for the pressure field? See e.g. 
https://github.com/MeltPoolDG/MeltPoolDG/blob/e359b21399af1394c5bffb455ed513daa81b8ac1/include/meltpooldg/simulations/rising_bubble/rising_bubble.hpp#L132

This is needed, since you only prescribe Dirichlet boundary conditions on the velocity field.

include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.hpp
```diff
@@ -17,7 +17,11 @@
...
  * International Journal of Multiphase Flow, 29(7), 1117-1135.
- * https://doi.org/10.1016/S0301-9322(03)0008
+ * https://doi.org/10.1016/S0301-9322(03)00084-3
+ *
+ * Balcázar, N., Rigola, J., Castro, J., & Oliva, A. (2016). A level-set model for thermocapillary
+ * motion of deformable fluid particles. International Journal of Heat and Fluid Flow, 62, 324-343.
+ * https://doi.org/10.1016/j.ijheatfluidflow.2016.09.015
```
mschreter, line 24: Maybe you can add a comment here that in the two papers the bubbles do not coalesce due to the mentioned reasons.

mschreter (commented):
Thank you Nils, looks good to me. I had some minor comments. Maybe you can address them, recreate the test output with the lowest verbosity level from adaflo (otherwise memory etc. is written, which is a rather sensitive measure for a test case), as Peter suggested. Then I am happy to merge.

include/meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.hpp
mschreter: However, we should keep this [Multiple Marker Level Set Method](https://reader.elsevier.com/reader/sd/pii/S0301932215001019?token=CB5B621B0AE50C51432ED7EE495CD28D26A6F865E496B732FDE439406E2D81091069ACDF07983056563EB0FE1DA7678C&originRegion=eu-west-1&originCreation=20210408114228) in mind, since it might be interesting to represent the powder particles

tests/thermo_capillary_two_droplets.with_adaflo_support=on.mpirun=4.output
```diff
@@ -39,28 +39,9 @@ t= 0.05      |vel|= 0
...
--- Statistics --   velocity  :   0.0148   0.0183   0.022  0.000914  2      0
--- Statistics --   div matrix:   0.0295   0.0336   0.0375  0.00168  0      2
--- Statistics --   pres mass :  0.00042 0.000499 0.000584 2.49e-05  2      0
--- Statistics --   pres Poiss:  0.00601  0.00603  0.00606 0.000302  0      2
-
  |velocity| = 5.76821458593279e-05
- |p| = 0.0366770191078899
-t= 0.1       |vel|= 0.0115550114438368
+ |p| = 0.0380818860671517
+t= 0.1       |vel|= 0.0115550114413756
```
nmuch, line 44: These number change slightly because of the new fix pressure boundary condition.

PR #126, opened 2021-04-09
Add jupyter notebook for two droplet simualtion parameters

nmuch:
This PR add a jupyter notebook for determining the concrete parameters from the dimensionless problem definition of the two thermo-capillary droplets simulation. For some parameters one must still choose arbitrary values.
Fyi @mschreter


mschreter:
Very nice, thank you! I particularly like that GitHub offers to run the file and automatically generate the preview. Maybe we can use ipynb for further documentations.

PR #128, opened 2021-04-12
Update convective Navier-Stokes-Formulation in case of evaporation

mschreter:
This PR is related to https://github.com/kronbichler/adaflo/pull/43 and updates primarily the test cases for two-phase-flow evaporation. In case of evaporation the convective term in the Navier-Stokes-Solver must be formulated in a convective way to obtain realistic results for the pressure. In addition, the following minor issues were adressed:

- for the output of density/viscosity an own DoFHandler is introduced
- `unit_tests` are now counted correctly by CMake


include/meltpooldg/flow/adaflo_wrapper.hpp
```diff
@@ -121,12 +126,24 @@ namespace MeltPoolDG::Flow
     {
       // clear constraints and setup hanging node constraints
       navier_stokes.distribute_dofs();
+      dof_handler_parameters.distribute_dofs(
+        FE_Q<dim>(navier_stokes.get_dof_handler_u().get_fe().tensor_degree()));
```
peterrum, line 130: This will not work with simplex ;)

include/meltpooldg/flow/adaflo_wrapper.hpp
```diff
@@ -121,12 +126,24 @@ namespace MeltPoolDG::Flow
...
       navier_stokes.initialize_data_structures();
     }
 
     void
     reinit_2()
+    {
+      constraints_parameters.clear();
+      constraints_parameters.reinit(scratch_data.get_locally_relevant_dofs(dof_index_parameters));
+      DoFTools::make_hanging_node_constraints(dof_handler_parameters, constraints_parameters);
+      constraints_parameters.close();
```
peterrum, line 142: I think we don't need a constraint matrix here. Just leave it empty.

peterrum: Then we would not need the new function. BTW: I should have given these functions more useful names^^

peterrum (approved):

unit_tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ DEAL_II_PICKUP_TESTS()
 
 
 SET(_n_tests "0")
-FILE(GLOB _tests ${_testexe}*.json)
+FILE(GLOB _tests ${_testexe}*.output)
```
peterrum, line 8: One more thing. This is not correct if we have multiple output files and/or the execution of the tests on some compile configurations. But I guess having having a number which is not zero is better than getting zero ;) In principal, what one might want to do is to loop over all targets and check if it starts with `unit_tests` or one let `DEAL_II_PICKUP_TESTS` return a list of tests.

include/meltpooldg/flow/adaflo_wrapper.hpp
peterrum: CI doesn't like this either!

unit_tests/CMakeLists.txt
mschreter: I agree but let's postpone this issue.

include/meltpooldg/flow/adaflo_wrapper.hpp
mschreter: CI got me ;-). I've fixed this issue.

include/meltpooldg/flow/adaflo_wrapper.hpp
mschreter: I've deleted the constraints since they are not really important in this case.

include/meltpooldg/flow/adaflo_wrapper.hpp
peterrum: I was quicker than CI :P

peterrum (approved):
I'll merge once CI is happy!

PR #137, opened 2021-04-22
Add badges

peterrum:
Add some nice badges:

Screenshot from 2021-04-22 12-39-42

We can move the badges to a better place once we rework the readme file (and maybe introduce a wiki).


mschreter:
Shall we still also use the GitHub CI routines?

peterrum:
> Shall we still use the GitHub CI routines?

I would suggest to keep both of them for now. Let's make a decision once we made some experiences.

PR #139, opened 2021-04-28
Introduce pvd file for time-dependent paraview data

mschreter:
This PR introduces the output of a `pvd`-file, enabling time-dependent (and not just time step dependent) output data. @nmuch @peterrum: FYI

### DIFF
--- a/include/meltpooldg/utilities/postprocessor.hpp
+++ b/include/meltpooldg/utilities/postprocessor.hpp
@@ -89,7 +89,7 @@ namespace MeltPoolDG
 
   private:
     void
-    write_paraview_files(const int n_time_step, const double time)
+    write_paraview_files(const unsigned int n_time_step, const double time)
     {
       DataOutBase::VtkFlags flags;
       if ((do_simplex == false) && (dim > 1))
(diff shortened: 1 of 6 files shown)


peterrum (approved):
Looks good to me. I have also tried it out locally. I have pushed two minor stylistic modifications. ~I'll merge once CI has passed.~

include/meltpooldg/utilities/postprocessor.hpp
```diff
@@ -37,6 +37,8 @@ namespace MeltPoolDG
 
     DataOut<dim> data_out;
 
+    std::vector<std::pair<double, std::string>> times_and_names;
```
peterrum, line 40: One more change. I have made this vector to class variable and removed the static. The static had the side effect that we could have a single `PostProcessor` class, since in the case of multiple post-processor classes all would append to the same vector, which does not make sense. Just image we have in the future multiple non-matching grids.

mschreter: Thank you Peter, very nice, the change makes sense. To be honest, I've copied this section from deal.II step 18.

peterrum: I know^^ Feel free to merge if you are happy!

PR #141, opened 2021-05-04
Add solidification to heat equation

nmuch:
Modify heat equation to handle temperature dependent parameters.

Introduce a solidus and liquidus temperature and solid material parameters.

Linear interpolate the materal parameters in the mushy zone.

Add solidification slab case and test.

This PR does not include latent heat effects. I'll make a follow-up PR for the apparent capacity method.


include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -229,29 +251,59 @@ namespace MeltPoolDG::Heat
               ls_vals.gather_evaluate(*level_set_as_heaviside, true, false);
             }
 
-          for (unsigned int q_index = 0; q_index < temp_vals.n_q_points; ++q_index)
+          if (data.solidification)
+            {
+              temperature_vals.reinit(cell);
+              temperature_vals.read_dof_values_plain(temperature);
+              temperature_vals.evaluate(true, true);
```
nmuch, line 258: For reading the velocity and level-set, we might want to use `read_dof_values_plain()` instead of `gather_evaluate()` because we use the plain values (constrained or not). Here, reading the temperature using `gather_evaulate()` would lead to unphysical behaviour if there were constrains. I guess we will run into similar problems if the velocity or level-set fields are constrained. Or am I mistaken?

nmuch: Nevermind, the issue I had was not related to `gather_evaluate()`. It should work fine here.

nmuch:
@mschreter as discussed I changed the solidification method to use the material.second properties for the liquid's properties, so we can rename the material phase structs accordingly in a follow-up: second -> liquid, first -> gas

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -18,39 +18,48 @@ namespace MeltPoolDG::Heat
...
+   *  dT_b^(n+1)     dt  \                            /
+   *                                                   Ω
+   *                1   /       d ρ |                          d c_p |                     \
+   *              + --  | N_a, -----| c_p^(n+1) T^(n+1) N_b + -------| ρ^(n+1) T^(n+1) N_b |
+   *                dt  \       d T |                           d T  |                     /
+   *                                |(n+1)                           |(n+1)                 Ω
+   *                  /                       d k |              \
+   *              +   | ∇N_a, k^(n+1) ∇N_b + -----| ∇T^(n+1) N_b |
+   *                  \                       d T |              /
+   *                                              |(n+1)          Ω
```
mschreter, line 58:
```suggestion
   *                  /                       d k |              \
   *              +   | ∇N_a, k^(n+1) ∇N_b + -----| T^(n+1) ∇N_b |
   *                  \                       d T |              /
   *                                              |(n+1)          Ω
```

mschreter: I think there is a mistake in the derivative. If this change also affects the implementation, please modify it there as well.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -69,7 +78,7 @@ namespace MeltPoolDG::Heat
    *
    *  q = α (T-T∞)
    *
-   * with the convection coeficient alpha.
+   * with the convection coefficient alpha.
```
mschreter, line 81:
```suggestion
   * with the convection coefficient denoted as α.
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -135,14 +144,27 @@ namespace MeltPoolDG::Heat
...
+      AssertThrow(
+        (material.first.conductivity > 0.0 && material.first.density > 0.0) || data.solidification,
+        ExcMessage("The material's conductivity and density must be greater than zero! Abort..."));
       AssertThrow(
         !level_set_as_heaviside ||
           (material.second.conductivity > 0.0 && material.second.density > 0.0),
         ExcMessage(
           "The secondary material's conductivity and density must be greater than zero! Abort..."));
+      AssertThrow(
+        !data.solidification || (material.solid.conductivity > 0.0 && material.solid.density > 0.0),
```
mschreter, line 156: Would it be more precise to have 
```suggestion
        data.solidification && (material.solid.conductivity > 0.0 && material.solid.density > 0.0),
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -135,14 +144,27 @@ namespace MeltPoolDG::Heat
...
+        ExcMessage(
+          "In case of solidification the solid's conductivity and density must be greater than zero! Abort..."));
+      AssertThrow(
+        !data.solidification ||
```
mschreter, line 160: same here?

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -135,14 +144,27 @@ namespace MeltPoolDG::Heat
...
+          (material.second.conductivity > 0.0 && material.second.density > 0.0),
+        ExcMessage(
+          "In case of solidification the liquid's (material.second) conductivity and density must be greater than zero! Abort..."));
+      AssertThrow(
+        !data.solidification || material.liquidus_temperature > material.solidus_temperature,
```
mschreter, line 165: same here?

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -229,29 +256,58 @@ namespace MeltPoolDG::Heat
...
-                                                  density);
+                {
+                  get_material_parameters_with_two_phase_flow(capacity,
+                                                              conductivity,
+                                                              density,
+                                                              ls_vals.get_value(q_index));
+                }
 
-              auto val = density * capacity * this->d_tau_inv * temp_vals.get_value(q_index);
+              if (data.solidification)
```
mschreter, line 275: What would happen if we have `level_set_as_heaviside` and `data.solidification` defined?

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -426,34 +524,62 @@ namespace MeltPoolDG::Heat
...
                     }
 
-                  temp_vals.submit_value(val, q_index);
+                  auto val_grad = conductivity * shape_vals.get_gradient(q_index);
 
-                  auto val_grad = conductivity * temp_vals.get_gradient(q_index);
+                  if (data.solidification)
+                    {
+                      val_grad += d_conductivity_dT * temperature_vals.get_gradient(q_index) *
+                                  shape_vals.get_value(q_index);
```
mschreter, line 573:
```suggestion
                      val_grad += d_conductivity_dT * temperature_vals.get_value(q_index) *
                                  shape_vals.get_gradient(q_index);
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -500,43 +633,70 @@ namespace MeltPoolDG::Heat
...
                           }
 
-                        temp_vals.submit_value(val, q_index);
+                        auto val_grad = conductivity * shape_vals.get_gradient(q_index);
 
-                        auto val_grad = conductivity * temp_vals.get_gradient(q_index);
+                        if (data.solidification)
+                          {
+                            val_grad += d_conductivity_dT * temperature_vals.get_gradient(q_index) *
+                                        shape_vals.get_value(q_index);
```
mschreter, line 686: Same change as in previous comment.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -705,16 +870,36 @@ namespace MeltPoolDG::Heat
...
+                }
 
               auto val = density * capacity *
                            (temp_vals.get_value(q_index) - temp_vals_old.get_value(q_index)) *
                            this->d_tau_inv -
                          heat_source_vals.get_value(q_index);
 
+              if (data.solidification)
+                {
+                  val -= (density_old * capacity_old - density * capacity) *
```
mschreter, line 899: This does not make sense. In case of solidification you would have multiply subtracted `density * capacity * temp_vals_old` (see lines 892-895).

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
 
   private:
     /*
-     * @return \c std::tuple with density, capacity and conductivity
+     * Determine material parameters (capacity, conductivity and density) for
+     * solidification/melting. In the mushy zone (between solidus and liquidus temperautre) the
```
mschreter, line 1023:
```suggestion
     * solidification/melting. In the mushy zone (between solidus and liquidus temperature) the
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
+     * material parameters are linearly interpolated between the solid's and fluid's values.
+     *
+     * @TODO, handle variable fluid parameters. ATM the fluid's parameters are assumed to be
+     * materal.second.<..>
```
mschreter, line 1027:
```suggestion
     * material.second.<..>
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
+     */
+    void
+    get_material_parameters_with_solidification(VectorizedArrayType &      capacity,
+                                                VectorizedArrayType &      conductivity,
+                                                VectorizedArrayType &      density,
+                                                const VectorizedArrayType &temperature) const
+    {
+      const auto unbound_mushy_zone_solid_fraction =
```
mschreter, line 1035:
```suggestion
/*
* Compute the solid fraction for a temperature between the liquidus and the solidus temperature. If the temperature is equal to the liquidus temperature, then the solid fraction is zero.  If the temperature is equal to the solidus temperature, then the solid fraction is one. 
*/
      const auto unbound_mushy_zone_solid_fraction =
```

mschreter (commented):
I'll finish for now, but still have to review certain parts of the code later.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
+        (material.liquidus_temperature - temperature) /
+        (material.liquidus_temperature - material.solidus_temperature);
+      const auto solid_fraction = compare_and_apply_mask<SIMDComparison::less_than>(
```
mschreter, line 1038: Couldn't you simply cut all `unbound_mushy_zone_solid_fraction` larger than 1 by 1 and smaller than 0 by 0?

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
+                                    temperature, material.liquidus_temperature, 1.0, 0.0) *
+                                    compare_and_apply_mask<SIMDComparison::greater_than>(
+                                      temperature, material.solidus_temperature, 1.0, 0.0) *
+                                    unbound_mushy_zone_solid_fraction +
+                                  compare_and_apply_mask<SIMDComparison::less_than_or_equal>(
+                                    temperature, material.solidus_temperature, 1.0, 0.0);
```
mschreter, line 1044: Could you consider a utility function in `utilityfunctions.hpp` for that?
```cpp
template <typename number>
    VectorizedArray<number>
    limit_to_bounds(const VectorizedArray<number> &in, const number& lower_limit, const number& upper_limit)
    {
      return compare_and_apply_mask<SIMDComparison::less_than> 
               (compare_and_apply_mask<SIMDComparison::greater_than>(in,
                                                                                                    upper_limit,
                                                                                                    upper_limit,
                                                                                                     in), 
                                    lower_limit, 
                                    lower_limit, 
                                    in);
    }
```
Then you could calculate the solid_fraction as follows:
```cpp
const auto solid_fraction =  (material.liquidus_temperature - limit_to_bounds(temperature,material.solidus_temperature, material.liquidus_temperature) /
        (material.liquidus_temperature - material.solidus_temperature);
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
-     * @return \c std::tuple with density, capacity and conductivity
+     * Determine material parameters (capacity, conductivity and density) for
+     * solidification/melting. In the mushy zone (between solidus and liquidus temperautre) the
+     * material parameters are linearly interpolated between the solid's and fluid's values.
+     *
+     * @TODO, handle variable fluid parameters. ATM the fluid's parameters are assumed to be
+     * materal.second.<..>
+     */
+    void
+    get_material_parameters_with_solidification(VectorizedArrayType &      capacity,
```
mschreter, line 1030: Since you always call `get_material_parameters` by `get_material_parameters_derivatives` would it make sense to have only one function, delivering both -- parameters and derivatives?
```suggestion
    get_solidifying_material_parameters_and_derivatives(VectorizedArrayType &      capacity,
     VectorizedArrayType &      conductivity,
     VectorizedArrayType &      density,
      VectorizedArrayType &      d_capacity_dT,
      VectorizedArrayType &      d_conductivity_dT,
      VectorizedArrayType &      d_density_dT,
      const VectorizedArrayType &temperature) const
```

include/meltpooldg/interface/parameters.hpp
```diff
@@ -837,6 +851,9 @@ namespace MeltPoolDG
           "heat variable properties over interface",
           heat.variable_properties_over_interface,
           "Set this parameter to smear the material properites over the two-phase interface.");
+        prm.add_parameter("heat solidification",
+                          heat.solidification,
+                          "Set this parameter to consider solidification.");
```
mschreter, line 856:
```suggestion
                          "Set this parameter to true to consider solidification.");
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -834,16 +1019,87 @@ namespace MeltPoolDG::Heat
...
+      const VectorizedArrayType &temperature) const
     {
-      // todo Input material parameters for two phase
+      const auto unbound_capacity_derivative =
+        (material.second.capacity - material.solid.capacity) /
+        (material.liquidus_temperature - material.solidus_temperature);
+      d_capacity_dT = compare_and_apply_mask<SIMDComparison::less_than>(
+                        temperature, material.liquidus_temperature, 1.0, 0.0) *
+                      compare_and_apply_mask<SIMDComparison::greater_than>(
+                        temperature, material.solidus_temperature, 1.0, 0.0) *
```
mschreter, line 1077: This coefficient is for all derivatives the same. Thus you could calculate it only once. Further, it could be nice to have a utility function for this purpose:
```cpp
template <typename number>
    VectorizedArray<bool>
    is_between(const VectorizedArray<number> &in, const number& lower_limit, const number& upper_limit)
    { ... }
```

include/meltpooldg/interface/parameters.hpp
```diff
@@ -1011,6 +1028,19 @@ namespace MeltPoolDG
         prm.add_parameter("material second viscosity",
                           material.second.viscosity,
                           "viscosity of the secondary material (level-set = 1)");
+        prm.add_parameter("material solid capacity",
+                          material.solid.capacity,
+                          "capacity of the solid");
```
mschreter, line 1033:
```suggestion
                          "capacity of the solid phase");
```

include/meltpooldg/interface/parameters.hpp
```diff
@@ -1011,6 +1028,19 @@ namespace MeltPoolDG
...
+        prm.add_parameter("material solid conductivity",
+                          material.solid.conductivity,
+                          "conductivity of the solid");
```
mschreter, line 1036:
```suggestion
                          "conductivity of the solid phase");
```

include/meltpooldg/interface/parameters.hpp
```diff
@@ -1011,6 +1028,19 @@ namespace MeltPoolDG
...
+        prm.add_parameter("material solid density", material.solid.density, "density of the solid");
```
mschreter, line 1037:
```suggestion
        prm.add_parameter("material solid density", material.solid.density, "density of the solid phase");
```

mschreter (changes requested):
Thank you @nmuch, I went through the PR and made some comments. If some of them are unclear, feel free to reach out any time.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -204,18 +226,23 @@ namespace MeltPoolDG::Heat
                       const VectorType &                    src,
                       std::pair<unsigned int, unsigned int> cell_range) const
     {
-      FECellIntegrator<dim, 1, number>   temp_vals(matrix_free, temp_dof_idx, this->quad_idx);
+      FECellIntegrator<dim, 1, number>   shape_vals(matrix_free, temp_dof_idx, this->quad_idx);
```
peterrum, line 229: Is there a particular reason for the renaming? If no, I would like to keep the old name, since the `FEEvaluation` object is linked to the ansatz/test functions of the temperature field.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -204,18 +226,23 @@ namespace MeltPoolDG::Heat
...
       FECellIntegrator<dim, 1, number>   ls_vals(matrix_free, ls_dof_idx, this->quad_idx);
+      FECellIntegrator<dim, 1, number> temperature_vals(matrix_free, temp_dof_idx, this->quad_idx);
 
       VectorizedArrayType capacity     = material.first.capacity;
       VectorizedArrayType conductivity = material.first.conductivity;
       VectorizedArrayType density      = material.first.density;
 
+      VectorizedArrayType d_capacity_dT     = 0.0;
+      VectorizedArrayType d_conductivity_dT = 0.0;
+      VectorizedArrayType d_density_dT      = 0.0;
```
peterrum, line 240: These can be moved into the cell loop.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -346,32 +407,62 @@ namespace MeltPoolDG::Heat
                   ls_vals.reinit(current_cell_index);
                   ls_vals.gather_evaluate(*level_set_as_heaviside, true, false);
                 }
+
+              if (data.solidification)
+                {
+                  temperature_vals.reinit(current_cell_index);
+                  temperature_vals.gather_evaluate(temperature, true, true);
+                }
```
peterrum, line 415: To me it looks like that `temperature_vals` is now responsible for the linearlization point of the Newton solver. Let's rename this `temp_old_vals` and `shape_vals` to `temp_vals`.

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -859,4 +1115,4 @@ namespace MeltPoolDG::Heat
         UtilityFunctions::interpolate(weight, material.first.density, material.second.density);
     }
   };
-} // namespace MeltPoolDG::Heat
+} // namespace MeltPoolDG::Heat
```
peterrum, line 1118: Could you add a new line.

peterrum (commented):
@nmuch I left some small comments.

Generally, I would suggest to create in a follow-up PR a free function that contains the body of the cell loop. Currently, we have more or less 3 versions of the same code, which was only an intermediate step to get the preconditioners running.

mschreter:
> @nmuch I left some small comments.
> 
> Generally, I would suggest to create in a follow-up PR a free function that contains the body of the cell loop. Currently, we have more or less 3 versions of the same code, which was only an intermediate step to get the preconditioners running.

That's a good idea. What signature of this free function do you have in mind?

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: No, the derivative I wrote is correct, at least that's what our other software uses. 
First I used your variant, but Newton will not converge. I'm still not quite sure why it doesn't.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: For the case we don't have solidification we don't need the materal.solid parameters. Your assert would throw every time we don't have solidification enabled. We would need a both or nothing bool operator.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: yes

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: yes

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: That case does not work yet. I'll look into a method to determine the material parameters for the combined case in a follow-up. Do you want me to add an assert for now?

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: So comment above. This should work fine.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: Yes. I did not change the default implementation for no solidification because the old material parameters are not set there. This tern subtracts `-density*capacity*temp_vals_old` again and add the correct `-density_old*capacity_old*temp_vals_old`. 

Maybe we should always compute the old material parameters, also if they don't change, so we don't have this weird line correcting the default equation. wdyt?

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: That's what I doing here. Is there a more efficient way I not aware or? I'm curious.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: Ah that's for the comment above, right?

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: I only call both for the tangent functions. For the residual we don't need the derivatives.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: I renamed this to not confuse it with `temperature_vals` with holds the actual temperature values in the tangent loops. I'm aware that `shape_vals` isn't correct here. I talked to @mschreter for a better name, but we didn't find one yet. 

The problem is that we need the total temperature values in the tangent loop to determine the material parameters, but `temp_vals` does not provide those as the name suggests.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: Now I get conflicting info 🤔 @mschreter preferred them to be outside the cell loop in a previous PR.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: I would not name is `temp_old_vals` because then i can be confused with the `temperature_old` values or the previous time step. It's the temperature values of the current time step but the previous iteration.

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: sure

nmuch:
Thank you @mschreter and @peterrum for your reviews. I'll implement the obvious changes right away.

> Generally, I would suggest to create in a follow-up PR a free function that contains the body of the cell loop.

Yes, that a good idea. In this PR I just implemented the new terms in the existing code. Having there sets of the same equation was quite annoying but restructuring the code is not the scope of this PR.

include/meltpooldg/heat/heat_transfer_operator.hpp
peterrum: Since it is the linearization point, we could call it `temp_lin_vals`.

peterrum:
> That's a good idea. What signature of this free function do you have in mind?

Nothing spectacular. Since the functions will not have some local variables in the scope like the FEEval instances these need to pass explicitly. I would make the following to a function:

https://github.com/MeltPoolDG/MeltPoolDG/blob/d05eb240de4b98f833f947855510003faf4d8024/include/meltpooldg/heat/heat_transfer_operator.hpp#L418-L428

and this one to a function:

https://github.com/MeltPoolDG/MeltPoolDG/blob/d05eb240de4b98f833f947855510003faf4d8024/include/meltpooldg/heat/heat_transfer_operator.hpp#L432-L456

in particular, I would make 

https://github.com/MeltPoolDG/MeltPoolDG/blob/d05eb240de4b98f833f947855510003faf4d8024/include/meltpooldg/heat/heat_transfer_operator.hpp#L402-L404

part of the latter function.

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Okay, then please ignore my previous comment :-).

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Exactöy-

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: I agree to the last point. The "old" parameters could be modified solely in case of solidification. Maybe you can also add a todo here that in case of two-phase-flow the "old" parameters would also change depending on the old level set values...

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: So please ignore the previous one :-)

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Yes, please add an assert so we know what to do in the future.

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Oh yes, I got it. Thank you for the explanation :-).

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: If its about "thread-safety" then we should include them inside the loop (which is necessary in view of introducing a free function for the cell operation anyhow). Sorry for the re-correction :-).

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: In my comments T^(n+1) are actually the nodal values (It would have been better to call them q_T^(n+1)). So to get the derivative one has to simply differentiate the shape functions and multiply by the nodal values, so I would prefer to modify the comment to 
```cpp
   *                  /                       d k |              \
   *              +   | ∇N_a, k^(n+1) ∇N_b + -----| T^(n+1) ∇N_b |
   *                  \                       d T |              /
   *                                              |(n+1)          Ω
   * ```
 The implementation is fine, since temperature_vales.get_gradient(q_index) corresponds to `T^(n+1) ∇N_b` and the multiplication with `shape_vals.get_value(q)` represents the multiplication with the Newton correction in the ith iteration dT^(i,n+1).

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: @nmuch: If you agree I could implement the cell operation in a free function in a follow-up PR.

nmuch: Sure, I'll make a PR with the tangent cell loop and move the declaration of the material variables in that PR.

mschreter:
I'll merge once CI is happy.

mschreter (approved):

PR #145, opened 2021-05-05
Introduce local cell and boundary operations for tangent in heat_transfer_operator

mschreter:
... as discussed. I've tested AMG|AMGReduced|ILU|ILUReduced|Diagonal|DiagonalReduced and the number of iterations remain the same.


peterrum (approved):
@mschreter Thanks a lot for cleaning up the code! I took the liberty to make some further simplifications!

Regarding:
> I've tested AMG|AMGReduced|ILU|ILUReduced|Diagonal|DiagonalReduced and the number of iterations remain the same.

Maybe we should think about testing all these preconditioners (at least that they converge and no assert is thrown)? I am bit hesitant in doing so: GitHub CI (debug mode) already needs an hour to complete. Maybe we should switch to release mode there also just like in the case of Travis?

mschreter:
Yes, I was also thinking about creating tests for the preconditioners. I can do that tomorrow. 

The debug mode has the advantage of catching potential asserts. Maybe we can identify the longer tests and decrease the run time first.

Issue #152, opened 2021-05-11
Postprocessing 


mschreter:
some brainstorming:

- [x] it should be possible to access the solution vectors within the application cases so that case-specific postprocessing routines can be implemented
- [x] add virtual method `SimulationBase::do_postprocessing(GenericDataOut & data_out)` and call it within the time loop
- [x] create a class `GenericDataOut` which allows to add all vectors, which can be addressed by its name (also add `DoFHandler`, ...). Vectors could be attached just like in the case of `DataOut`:
```cpp
  generic_data_out.add_data_vector(scratch_data->get_dof_handler(velocity_dof_idx),
    advection_velocity, std::vector<std::string>(dim, "velocity"), vector_component_interpretation );
```
- [x] create a separate ` ParaviewPostprocessor` class, which fills `DataOut` based on `GenericDataOut`

PR #154, opened 2021-05-12
Do not use DataOut::attach_dof_handler()

peterrum:
... as a preparation of #152. In a follow-up PR, I would like to remove the direct usage of `DataOut`.


mschreter (approved):
Fine for me. Feel free to merge!

PR #155, opened 2021-05-12
Instantiate templates

mschreter:


peterrum:
> No it is not and takes some time :-). Would you agree if I finish in the first PR the instantiations and subsequently move the implementation to the cpp-files in problemwise PR?

Fine with me, in the case that there are no multiple instantiation errors :)

CMakeLists.txt
```diff
@@ -33,34 +47,60 @@ ENDIF()
...
 INCLUDE_DIRECTORIES(include ${CMAKE_CURRENT_BINARY_DIR}/include)
 
-PROJECT(MeltPoolDG)
+file(GLOB_RECURSE sources "${CMAKE_SOURCE_DIR}/include/*.hpp" "${CMAKE_SOURCE_DIR}/source/*.cpp" "${CMAKE_SOURCE_DIR}/source/**/*.cpp")
+
+ADD_LIBRARY(${libName}
+            STATIC
+            ${sources})
+
+DEAL_II_SETUP_TARGET(${libName})
```
peterrum, line 84: I am nor sure if this line is needed.

CMakeLists.txt
```diff
@@ -33,34 +47,60 @@ ENDIF()
...
+
+################################
+# create executable
+################################
 
-SET(projectName "meltpooldg")
+SET(projectName "meltpooldg-bin")
```
peterrum, line 90: What about calling this `meltpooldg-app`?

mschreter: I've introduced a renaming to `meltpooldg`.

CMakeLists.txt
mschreter: Jap, it's needed.

mschreter:
Locally I don't have any indent problems...

peterrum:
This is odd. For some reason it thinks that `#include <meltpooldg/utilities/utilityfunctions.hpp>` is written twice. What about `// clang-format off`. I will close my eyes in this instance^^

mschreter:
Thanks, then this PR would be ready from my side :-).

peterrum:
> Thanks, then this PR would be ready from my side :-).

I have removed it again ;) The problem was that clang-format is not run on the branch but on the branch merged into master. And in that version indeed the include was twice!

peterrum (approved):

PR #157, opened 2021-05-12
Add terms for convection with temperature-dependent material parameters to heat operator

nmuch:
The second order temperature derivative is zero for now but this will change in a follow-up.

### DIFF
--- a/include/meltpooldg/heat/heat_transfer_operator.hpp
+++ b/include/meltpooldg/heat/heat_transfer_operator.hpp
@@ -22,57 +22,67 @@ namespace MeltPoolDG::Heat
    * equation with temperature dependent material properties:
    * ρ^(n) = ρ(T^(n)), c_p^(n) = c_p(T^(n)), k^(n) = k(T^(n))
    *
-   *                    1   /                                                               \
-   *  R(T_b^(n+1)) =   ---  | N_a, ( N_b ρ^(n+1) c_p^(n+1) T_b^(n+1) - ρ^(n) c_p^(n) T^(n)) |
-   *                   dt   \                                                               /
-   *                                                                                         Ω
-   *                 /                                         \
-   *               + | N_a, ρ^(n+1) c_p^(n+1) u ∇N_b T_b^(n+1) |
-   *                 \                                         /
-   *                                                            Ω
-   *                 /                                          \
-   *               + | N_a, ρ^(n+1) c_p^(n+1) N_b T_b^(n+1) ∇·u |    (this term is not yet considered @todo)
-   *                 \                                          /
-   *                                                            Ω
+   *                  1  /                                                               \
+   *  R(T_b^(n+1)) = --- | N_a, ( ρ^(n+1) c_p^(n+1) N_b T_b^(n+1) - ρ^(n) c_p^(n) T^(n)) |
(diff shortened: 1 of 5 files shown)


include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -827,6 +854,20 @@ namespace MeltPoolDG::Heat
...
             }
 
+          if (velocity && data.solidification)
+            {
+              get_material_parameter_second_derivatives_with_solidification(
+                d2_capacity_dT2, d2_density_dT2, temp_lin_vals.get_value(q_index));
+              val += (d_capacity_dT * density + d_density_dT * capacity) *
+                     (2.0 * temp_lin_vals.get_value(q_index) * temp_vals.get_gradient(q_index) +
+                      temp_vals.get_value(q_index) * temp_lin_vals.get_gradient(q_index)) *
+                     velocity_vals.get_value(q_index);
```
mschreter, line 864: Couldn't you simply omit the second line?
```suggestion
                     (2.0 * temp_lin_vals.get_value(q_index) * temp_vals.get_gradient(q_index) *
                     velocity_vals.get_value(q_index);
```

include/meltpooldg/heat/heat_transfer_operator.hpp
```diff
@@ -827,6 +854,20 @@ namespace MeltPoolDG::Heat
...
+              val += (d2_capacity_dT2 * density + 2.0 * d_capacity_dT * d_density_dT +
+                      d2_density_dT2 * capacity) *
+                     temp_lin_vals.get_value(q_index) * temp_lin_vals.get_gradient(q_index) *
+                     velocity_vals.get_value(q_index);
```
mschreter, line 868: Is here the matrix-vector product multiplication missing?
```suggestion
                     * temp_vals.get_value(q_index);
```

mschreter (approved):
Thanks, looks fine for me. I had just two minor questions which might become more important as soon as you introduce the higher-order parameter interpolation :-).

include/meltpooldg/heat/heat_transfer_operator.hpp
nmuch: When deriving the consistent tangent from the residual, the product rule results in two identical terms. Here I used the factor 2 instead of writing down the term twice. I already updated the Theory Manual if you want to to have a look at the derivation.

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: You need it since you directly pass the matrix-vector product into `dst`.

nmuch: Sure, I'll add it

include/meltpooldg/heat/heat_transfer_operator.hpp
mschreter: Okay, now I can follow :-). The expression should be correct.

mschreter:
Thanks @nmuch I'll merge this PR now so that we can continue to work on the open topics. As discussed, we need to have a closer look into the correctness of the tangent (see #159).

PR #158, opened 2021-05-12
Implement GenericDataOut

peterrum:
references #152

This class should give us the flexibility to access vector by a simply providing the name. Magdalena could you take over from here and pass `GenericDataOut` somehow to the simulation classes. Thx!


peterrum:
BTW: I haven't tested the new class. Magdalena, could you run one case if it indeed works. But else it should be ready from my side!

mschreter:
@peterrum, thanks a lot for your efforts, I am happy to take over now. I've had a look at the output of one simulation -- no problems here :-).

Issue #151, opened 2021-05-11
Instantiate template classes in source files


mschreter:
... and compile parts into a library.

mschreter:
source/advection_diffusion:
- [ ] advection_diffusion_adaflo_wrapper.cpp  
- [ ] advection_diffusion_operation.cpp  
- [ ] advection_diffusion_operator.cpp  
- [ ] advection_diffusion_problem.cpp

source/curvature:
- [ ] curvature_operation_adaflo_wrapper.cpp  
- [ ] curvature_operation.cpp  
- [ ] curvature_operator.cpp

source/evaporation:
- [ ] evaporation_operation.cpp 
- [ ] evaporation_operation_marching_cube.cpp

source/flow:
- [ ] adaflo_wrapper.cpp  
- [ ] adaflo_wrapper_parameters.cpp 
- [ ] flow_base.cpp  surface_tension_operation.cpp  
- [ ] two_phase_flow_problem.cpp

source/heat
- [ ] heat_transfer_operation.cpp  
- [ ] heat_transfer_operator.cpp  
- [ ] heat_transfer_preconditioner.cpp  
- [ ] heat_transfer_problem.cpp  
- [ ] laser.cpp  
- [ ] laser_heat_source_gusarov.cpp

interface:
- [ ] boundaryconditions.cpp  
- [ ] fieldconditions.cpp  
- [ ] operator_base.cpp  
- [ ] parameters.cpp  
- [ ] periodic_boundary_conditions.cpp  
- [ ] problem_selector.cpp  
- [ ] scratch_data.cpp

source/level_set:
- [ ] level_set_operation.cpp  level_set_operator_with_phase_change.cpp  level_set_problem.cpp

source/melt_pool:
- [ ] melt_pool_operation.cpp  recoil_pressure_operation.cpp

source/normal_vector:
- [ ] normal_vector_operation_adaflo_wrapper.cpp  normal_vector_operation_base.cpp  normal_vector_operation.cpp  normal_vector_operator.cpp

source/reinitialization:
- [ ] olsson_operator.cpp  reinitialization_operation_adaflo_wrapper.cpp  reinitialization_operation_base.cpp  reinitialization_operation.cpp  reinitialization_problem.cpp

source/simulations:
- [ ] simulation_selector.cpp
@peterrum: How should we proceed with the simulations? Also both header and source files?

source/utilities:
[...]

PR #176, opened 2021-05-20
Precompile Interface -- Part 1: operator base, boundary and field conditions

mschreter:
references #151 

~todo: I have to work on a linking error regarding operator base.~


source/interface/operator_base.cpp
```diff
@@ -2,5 +2,100 @@
...
+  OperatorBase<number, DoFVectorType, SrcRhsVectorType>::reset_indices(
+    const unsigned int dof_idx_in,
+    const unsigned int quad_idx_in)
+  {
+    this->dof_idx  = dof_idx_in;
+    this->quad_idx = quad_idx_in;
+  }
+
+  template <typename number, typename DoFVectorType, typename SrcRhsVectorType>
+  template <int dim>
```
peterrum, line 24: The undefined reference is due to the additional template parameter here and below. While looking at the diff, I think it would be good to get rid of the `typename DoFVectorType, typename SrcRhsVectorType` maybe by introducing some free functions?

mschreter: For the moment I simply added the `dim` as a class template argument. I hope that's fine with you.

peterrum (approved):

PR #174, opened 2021-05-20
Update evaporation and Stefan problem 1

mschreter:
This PR includes:

- Modification of Stefan problem 1 (including test cases)
- Modification of the heat operator --> take into account evaporative heat flux also in case of constant evaporative mass flux determined at level set 0

PR #177, opened 2021-05-20
Precompile the evaporation and melt_pool folder

peterrum:
references #151


mschreter:
Can we merge this PR after #174? The tests of #174 should now pass.

PR #178, opened 2021-05-20
Precompile interface: parameters + PBC

mschreter:
references #151


peterrum (approved):
I didn't know that there are so many parameters by now. It started with a hand full...

mschreter:
> I didn't know that there are so many parameters by now. It started with a hand full...

:-). We try to be "user-friendly" ;-).

PR #188, opened 2021-05-21
Draft for accessing result vectors

mschreter:
This PR references #152. @peterrum: If this is what you thought about, I'll continue by integrating the user output also in the other problems.

@nmuch: This way, output vectors are available in the `Simulation` files. The goal is to directly obtain e.g. a value over time in table form or any other postprocessed variable you are interested in. This should save us in certain cases from opening Paraview at all.

### DIFF
--- a/include/meltpooldg/interface/simulationbase.hpp
+++ b/include/meltpooldg/interface/simulationbase.hpp
@@ -10,6 +10,7 @@
 #include <meltpooldg/interface/fieldconditions.hpp>
 #include <meltpooldg/interface/parameters.hpp>
 #include <meltpooldg/interface/periodic_boundary_conditions.hpp>
+#include <meltpooldg/utilities/generic_data_out.hpp>
 // c++
 #include <memory>
 
@@ -60,6 +61,17 @@ namespace MeltPoolDG
           " this->triangulation = std::make_shared<parallel::distributed::Triangulation<dim>>(this->mpi_communicator); "));
       set_boundary_conditions();
       set_field_conditions();
+    }
+
+    /**
+     * This function may be overwritten for output purposes in your simulation case.
+     * It is called after every time step.
+     */
+    virtual void
+    do_postprocessing(const GenericDataOut<dim> &generic_data_out)
+    {
+      (void)generic_data_out;
+      // do nothing default
     }
 
     /**
--- a/include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
+++ b/include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
@@ -142,8 +142,9 @@ namespace MeltPoolDG::Simulation::AdvectionDiffusion
         for (const auto &name : std::get<2>(entry))
           {
             pcout << " * " << std::setw(20) << name << " Max-norm: " << std::setprecision(5)
-                  << std::setw(10) << std::get<1>(entry)->linfty_norm()
-                  << " number of dofs: " << std::get<0>(entry)->n_dofs() << std::endl;
+                  << std::setw(10) << generic_data_out.get_vector(name).linfty_norm()
+                  << " number of dofs: " << generic_data_out.get_dof_handler(name).n_dofs()
+                  << std::endl;
             break;
           }
       pcout << "---------------------------------------------" << std::endl;
--- a/include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
+++ b/include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
@@ -1,0 +1,158 @@
+#pragma once
+#include <deal.II/base/function.h>
+#include <deal.II/base/point.h>
+#include <deal.II/base/tensor_function.h>
+
+#include <deal.II/grid/grid_generator.h>
+
+#include <deal.II/lac/vector.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+#include <meltpooldg/interface/simulationbase.hpp>
+#include <meltpooldg/simulations/advection_diffusion/advection_diffusion.hpp>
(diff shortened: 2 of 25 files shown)


include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
```diff
@@ -0,0 +1,158 @@
...
+          {
+            pcout << " * " << std::setw(20) << name << " Max-norm: " << std::setprecision(5)
+                  << std::setw(10) << std::get<1>(entry)->linfty_norm()
+                  << " number of dofs: " << std::get<0>(entry)->n_dofs() << std::endl;
+            break;
+          }
+      pcout << "---------------------------------------------" << std::endl;
+      pcout << "    End of user defined postprocessing" << std::endl;
+      pcout << "---------------------------------------------" << std::endl;
+    }
```
mschreter, line 152: This is essentially how it looks like to access the output. One could think about integrating more handy access functions into `GenericDataOut`.

source/advection_diffusion/advection_diffusion_problem.cpp
```diff
@@ -287,6 +291,12 @@ namespace MeltPoolDG::AdvectionDiffusion
                                std::vector<std::string>(dim, "velocity"),
                                vector_component_interpretation);
     };
+
+    if (do_attach_vectors)
+      attach_output_vectors(base_in->generic_data_out);
+
+    base_in->do_postprocessing();
+
     post_processor->process(time_step, attach_output_vectors, current_time);
```
peterrum, line 300: Personally, I think all the post processing could be moved to `Simulation`/`SimulationBase`. `SimulationBase` could contain the basic post processing call (like Paraview output) and the derived classes would override that function.

include/meltpooldg/interface/simulationbase.hpp
```diff
@@ -376,6 +387,7 @@ namespace MeltPoolDG
     const MPI_Comm                                mpi_communicator;
     Parameters<double>                            parameters;
     std::shared_ptr<Triangulation<dim, spacedim>> triangulation;
+    GenericDataOut<dim>                           generic_data_out;
```
peterrum, line 390: I would pass this to the `do_postprocessing()` function call (i.e. not make it to a class member). I am wondering if we don't need more information like current timestep or current time just like in the case of the current `PostProcessor` class. One option would to pass the info directly to the function or make the info part of `GenericDataOut`.

peterrum (commented):
Overall, looks very good. I left some first comments. Feel free to ignore them!

include/meltpooldg/interface/simulationbase.hpp
mschreter: That's true. I've removed the class member. Extending `GenericDataOut` and moving the postprocessing to the base class will be part of a follow-up PR.

mschreter:
NOTE: I had to exclude this assert to pass the (simplex-related) tests in debug mode:

https://github.com/mschreter/MeltPoolDG/blob/c363ebfc74a7e7145f65d737295e8c5fd5849e0e/include/meltpooldg/utilities/utilityfunctions.hpp#L69-L72

include/meltpooldg/simulations/advection_diffusion/advection_diffusion_user_output.hpp
```diff
@@ -0,0 +1,158 @@
...
+      dealii::ConditionalOStream pcout(std::cout,
+                                       Utilities::MPI::this_mpi_process(this->mpi_communicator) ==
+                                         0);
+
+      pcout << "---------------------------------------------" << std::endl;
+      pcout << "    Starting user defined postprocessing" << std::endl;
+      pcout << "---------------------------------------------" << std::endl;
+      pcout << "Accessible vectors:" << std::endl;
+      for (const auto &entry : generic_data_out.entries)
+        for (const auto &name : std::get<2>(entry))
```
peterrum, line 142: In the long term, I would prefer if we would access these class fields indirectly. In particular, a function would be nice that returns the vector with the correct name.

include/meltpooldg/utilities/utilityfunctions.hpp
```diff
@@ -67,9 +67,9 @@ namespace MeltPoolDG
       FECellIntegrator<dim, n_components, double> fe_eval(matrix_free, dof_idx, quad_idx);
 
       // @todo: replace n_q_points_1D argument completely from fe_eval?
-      AssertThrow(fe_eval.n_q_points == std::pow(n_q_points_1D, dim),
-                  ExcMessage("The number of quadrature points in 1D must comply with the number of "
-                             "quadrature points of the cell operation."));
+      // AssertThrow(fe_eval.n_q_points == std::pow(n_q_points_1D, dim),
+      // ExcMessage("The number of quadrature points in 1D must comply with the number of "
+      //"quadrature points of the cell operation."));
```
peterrum, line 72: I would prefer to keep the assert here. The function just returns non-sense for simplex (I am surprised that you don't get a segfault). Extending this function to simplex is a bit more involved... One might need `FE_DGQArbitraryNodes` also for simplex. I would suggest to call this function only if we don't use simplex. As far as I see it we always have to relevant info around.

peterrum (approved):
Feel free to merge (once you re-enabled the assert), if you don't want to add additional features now.

include/meltpooldg/utilities/utilityfunctions.hpp
mschreter: I am estonished that it did not hit us before. Anyway, I've added an assert if we have a non-hex-type-mesh.

peterrum (approved):

mschreter:
> I wouldn't introduce a new map. Simply search in the vector above. If you want I can implement it in a follow-up PR :)

Okay :-). I thought a map might be handy :-).

peterrum (approved):
I'll merge once CI is happy!

PR #192, opened 2021-05-21
Stefan's problem 1: tabular output of values at level set zero to file

mschreter:
This PR enables the direct output of data at level set zero for Stefan's problem 1 to a text file.

### DIFF
--- a/include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
+++ b/include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
@@ -1,10 +1,13 @@
 #pragma once
 // deal-specific libraries
 #include <deal.II/base/function.h>
+#include <deal.II/base/mpi.h>
+#include <deal.II/base/mpi.templates.h>
 #include <deal.II/base/point.h>
 #include <deal.II/base/tensor_function.h>
 
 #include <deal.II/grid/grid_generator.h>
+#include <deal.II/grid/grid_tools.h>
 #include <deal.II/grid/tria.h>
 
 #include <deal.II/lac/vector.h>
@@ -140,6 +143,10 @@ namespace MeltPoolDG::Simulation::StefansProblem1WithFlowAndHeat
 
     {
       this->set_parameters();
+      if (Utilities::MPI::this_mpi_process(mpi_communicator) == 0)
+        file_level_set_contour.open(this->parameters.paraview.directory + "/" +
+                                    this->parameters.paraview.filename +
+                                    "_level_set_contour_over_time");
     }
 
     void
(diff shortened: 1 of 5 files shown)


include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
```diff
@@ -317,7 +324,58 @@ namespace MeltPoolDG::Simulation::StefansProblem1WithFlowAndHeat
...
+                         vertices,
+                         cells);
+
+              auto vertices_all = Utilities::MPI::all_reduce<std::vector<Point<dim>>>(
+                vertices,
+                this->mpi_communicator,
+                [](const std::vector<Point<dim>> &a, const std::vector<Point<dim>> &b) {
+                  auto result = a;
+                  result.insert(result.end(), b.begin(), b.end());
+                  return result;
```
peterrum, line 355: To me it looks like that you need just one value for outputting, so why not simply writing:
```suggestion
                  return a.size() == 1 ? a : b;
```

include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
```diff
@@ -317,7 +324,58 @@ namespace MeltPoolDG::Simulation::StefansProblem1WithFlowAndHeat
...
+                generic_data_out.get_dof_handler("level_set").get_fe(),
+                1);
+
+              mc.process(generic_data_out.get_dof_handler("level_set"),
+                         generic_data_out.get_vector("level_set"),
+                         0.0,
+                         vertices,
+                         cells);
+
+              auto vertices_all = Utilities::MPI::all_reduce<std::vector<Point<dim>>>(
```
peterrum, line 349:
```suggestion
              auto vertices_all = Utilities::MPI::all_reduce<std::vector<Point<dim>>>(
```

peterrum, line 349: By the way, let me introduce a `Utilities::MPI::all_reduce` function. This is what you actually need here.

peterrum (commented):
I see you want to be a teacher's pet ;) Efficiently in (actual) 5 lines of code, you have used 3-4 new features that I have contributed to deal.II in the last months. I like that!

Conceptually, what you have written is correct. However let me suggest an alternative approach. Before doing that let me give the reason. You perform the following 3 steps: 1) compute points on the interface (LS = 0) with MCA, 2) collect the points on rank 0, and finally 3) evaluate the temperature field at these points (with RPE/VT). This is the right approach for non-matching grids (and as a result it also works for matching grids as here), e.g., the LS and heat are solved on different meshes, i.e., the cells of the LS and heat `DoFHandler` do not correspond. In the extreme case, the coarse grid might be completely different, the cells might be refined differently, and the triangulations might be partitioned completely differently among processes. To support these cases, PRE needs to perform an (expensive) search with two steps: step 1) find the processes that might have a cell containing the evaluation point (based on bounding boxes) and 2) actually find the cell on that process.

However, in the use case here the cells match so that you could use the same approach as in `EvaporationOperationMarchingCube::compute_mass_balance_source_term_sharp()`. So my suggestion would be to take an approach similar to there. 1) Loop over all cells and fill a vector  (of points + LS) right away, and 2) collect the result on rank 0 and print to file.

My suggestion would be to generalize the code in `EvaporationOperationMarchingCube::compute_mass_balance_source_term_sharp()` so that one could pass in a lambda that determines what to do with the resulting points!?

_Note: RPE/VT will be definitely needed in following postprocessor functions :+1:_

include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
mschreter: > I see you want to be a teacher's pet ;) Efficiently in (actual) 5 lines of code, you have used 3-4 new features that I have contributed to deal.II in the last months. I like that!
> 
> Conceptually, what you have written is correct. However let me suggest an alternative approach. Before doing that let me give the reason. You perform the following 3 steps: 1) compute points on the interface (LS = 0) with MCA, 2) collect the points on rank 0, and finally 3) evaluate the temperature field at these points (with RPE/VT). This is the right approach for non-matching grids (and as a result it also works for matching grids as here), e.g., the LS and heat are solved on different meshes, i.e., the cells of the LS and heat `DoFHandler` do not correspond. In the extreme case, the coarse grid might be completely different, the cells might be refined differently, and the triangulations might be partitioned completely differently among processes. To support these cases, PRE needs to perform an (expensive) search with two steps: step 1) find the processes that might have a cell containing the evaluation point (based on bounding boxes) and 2) actually find the cell on that process.
> 
> However, in the use case here the cells match so that you could use the same approach as in `EvaporationOperationMarchingCube::compute_mass_balance_source_term_sharp()`. So my suggestion would be to take an approach similar to there. 1) Loop over all cells and fill a vector (of points + LS) right away, and 2) collect the result on rank 0 and print to file.
> 
> My suggestion would be to generalize the code in `EvaporationOperationMarchingCube::compute_mass_balance_source_term_sharp()` so that one could pass in a lambda that determines what to do with the resulting points!?
> 
> _Note: RPE/VT will be definitely needed in following postprocessor functions +1_

To be honest, I did not put a lot of effort by improving the performance but rather using the new user-friendly functions that are really nice. So thanks for that :-). However, I agree that we might have a utility function for the evaluation along interface cells. So I've introduced one in `utilityfunctions.hpp`. I am using the latter now in `EvaporationOperationMarchingCube::compute_mass_balance_source_term_sharp()` as well as in `SimulationStefansProblem1WithFlowAndHeat::do_postprocessing`. Hope this complies with what you were thinking of.

include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
mschreter: ?

mschreter: ?

include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
mschreter: > To me it looks like that you need just one value for outputting, so why not simply writing:

Unfortunately, I end up with a segmentation fault so I stick to the old version :/.

peterrum: > Unfortunately, I end up with a segmentation fault so I stick to the old version :/.

Probably because the invariant `a.size() == 1` is not fulfilled. A solution would be `vertices.resize(std::min(1, vertices.size()));`.

include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp
peterrum: I meant `Utilities::MPI::reduce()`^^

peterrum:
```suggestion
              const auto vertices_all = Utilities::MPI::all_reduce<std::vector<Point<dim>>>(
```

peterrum (approved):
> Hope this complies with what you were thinking of.

Jup! Thanks!

PR #196, opened 2021-05-23
Fix deprecation warnings

peterrum:
due to https://github.com/dealii/dealii/pull/12289

similar to https://github.com/kronbichler/adaflo/pull/46


mschreter:
Thanks!

PR #198, opened 2021-05-25
Resolve: Fix parallel bug related to solidification in HeatTransferOperator

### LINKED ISSUE #197: Fix parallel bug related to solidification in HeatTransferOperator
The bug was the first time observed in PR #191. As a temporary workaround the affected test case `unidirectional_heat_transfer_with_two_phase_and_solidification.json` is run using only 1 process. Currently, it is observed that by using n+1 processes the convergence of the Newton is getting worse until divergence.

nmuch:
After updating deal.II I cannot reproduce the bug #197 and the bug does not come up for multiple MPI processes.

This PR's only change is to make the test `unidirectional_heat_transfer_with_two_phase_and_solidification` run with 4 MPI processes. Let's see if Github CI passes now.

Closes #197


mschreter:
That's good news :-). Although I do not understand at all what the reason for this could be. Do you have any ideas or any lines under suspision? So let's wait until CI is happy.

nmuch:
I don't know what fixed the bug. 
The deal.II version that I used when having this bug was commit 6735b5037e1aacfff790af8fa44ffcaf3ca89c93 which merged PR dealii/dealii#12249. I could not identify a recent deal.II PR that may be related.

PR #200, opened 2021-05-25
Thermo-capillary droplet in 3D

nmuch:
Here are the changes I made to run thermo-capillary droplet in 3D. Also I added the input file.

Fyi @mschreter


mschreter:
We have reached PR #200  :tada:.

PR #162, opened 2021-05-19
Fix heat equation for temperature dependent params

nmuch:
This PR fixes the heat equation for temperature dependent material parameters (during solidification). 
@mschreter When applying the mass balance assumption you sent me, a lot of terms cancel out.
With the consistent tangent the Newton shows quadratic convergence.

Also, in the mushy zone a cubic material parameter interpolation is used to have a continuous derivative. at the solidus and liquidus temperature.

PR #201, opened 2021-05-26
Melt front propagation

nmuch:
This PR adds a simple melt front propagation simulation example: A solid piece of metal is heated with a laser heat source and melts. The solid and liquid phases have different thermal properties.
Screenshot from 2021-05-26 10-24-34

A couple of changes come with this addition:
- add user input for Gusarov laser
- restructure `LaserHeatSourceGusarov` so it can be used in a `dealii::Function`
- add a source field function and compute heat source vector

### DIFF
--- a/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
+++ b/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
@@ -39,7 +39,6 @@ namespace MeltPoolDG::Heat
   private:
     using VectorType = LinearAlgebra::distributed::Vector<double>;
 
-    const ScratchData<dim> &scratch_data;
     /*
      *  Parameters for the Gusarov model
      */
@@ -47,37 +46,37 @@ namespace MeltPoolDG::Heat
 
     const double lambda; // optical thickness (-)
 
-    const unsigned int temp_dof_idx;
-
   public:
-    LaserHeatSourceGusarov(const ScratchData<dim> &       scratch_data_in,
-                           LaserData<double>::GusarovData gusarov_data_in,
-                           unsigned int                   temp_dof_idx_in);
+    LaserHeatSourceGusarov(const LaserData<double>::GusarovData &gusarov_data_in);
 
     void
-    compute_volumetric_heat_source(VectorType &      heat_source_vector,
-                                   double            laser_power,
-                                   const Point<dim> &laser_position,
-                                   bool              zero_out = true);
-
-  private:
-    /**
-     * Equation (26) corrected to match Figure 5 of Gusarov et al. (2009)
-     */
-    double
-    power_density(double radius, double power);
-
-    /**
-     * Analytical derivative of equation (20)
-     */
-    double
-    dq_dxi(double xi);
+    compute_volumetric_heat_source(VectorType &            heat_source_vector,
+                                   const ScratchData<dim> &scratch_data,
+                                   unsigned int            temp_dof_idx,
+                                   double                  laser_power,
+                                   const Point<dim> &      laser_position,
+                                   bool                    zero_out = true);
 
     /**
      * volumetric heat source; The z-axis (= axis of the laser beam) is assumed to correspond to
      * negative dim-1 coordinate.
      */
     double
-    local_heat_source(const Point<dim> &position, const Point<dim> &laser_position, double power);
+    get_local_heat_source(const Point<dim> &position,
+                          const Point<dim> &laser_position,
+                          double            power) const;
+
+  private:
+    /**
+     * Equation (26) corrected to match Figure 5 of Gusarov et al. (2009)
+     */
+    double
+    power_density(double radius, double power) const;
+
+    /**
+     * Analytical derivative of equation (20)
+     */
+    double
+    dq_dxi(double xi) const;
   };
 } // namespace MeltPoolDG::Heat
--- a/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
+++ b/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
@@ -49,6 +49,9 @@ namespace MeltPoolDG::Heat
   public:
     LaserHeatSourceGusarov(const LaserData<double>::GusarovData &gusarov_data_in);
 
+    /**
+     * Compute a DoF vector of the heat source from the Gusarov laser model.
+     */
     void
     compute_volumetric_heat_source(VectorType &            heat_source_vector,
                                    const ScratchData<dim> &scratch_data,
@@ -62,9 +65,9 @@ namespace MeltPoolDG::Heat
      * negative dim-1 coordinate.
      */
     double
-    get_local_heat_source(const Point<dim> &position,
-                          const Point<dim> &laser_position,
-                          double            power) const;
+    local_compute_volumetric_heat_source(const Point<dim> &position,
+                                         const Point<dim> &laser_position,
+                                         double            power) const;
 
   private:
     /**
@@ -74,9 +77,12 @@ namespace MeltPoolDG::Heat
     power_density(double radius, double power) const;
 
     /**
-     * Analytical derivative of equation (20)
+     * Analytical derivative of equation (20) of Gusarov et al. (2009)
      */
     double
     dq_dxi(double xi) const;
+
+    const double a;
+    const double D;
   };
 } // namespace MeltPoolDG::Heat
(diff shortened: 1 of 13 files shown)


include/meltpooldg/heat/laser_heat_source_gusarov.hpp
```diff
@@ -39,45 +39,44 @@ namespace MeltPoolDG::Heat
...
 
-  private:
     /**
-     * Equation (26) corrected to match Figure 5 of Gusarov et al. (2009)
+     * volumetric heat source; The z-axis (= axis of the laser beam) is assumed to correspond to
+     * negative dim-1 coordinate.
      */
     double
-    power_density(double radius, double power);
+    get_local_heat_source(const Point<dim> &position,
```
mschreter, line 65: Maybe we can rename it to `local_compute_volumetric_heat_source`.

include/meltpooldg/heat/laser_heat_source_gusarov.hpp
```diff
@@ -39,45 +39,44 @@ namespace MeltPoolDG::Heat
...
+     * Equation (26) corrected to match Figure 5 of Gusarov et al. (2009)
      */
     double
-    dq_dxi(double xi);
+    power_density(double radius, double power) const;
 
     /**
-     * volumetric heat source; The z-axis (= axis of the laser beam) is assumed to correspond to
-     * negative dim-1 coordinate.
+     * Analytical derivative of equation (20)
```
mschreter, line 77:
```suggestion
     * Analytical derivative of equation (20) of Gusarov et al. (2009)
```

include/meltpooldg/heat/laser_heat_source_gusarov.hpp
```diff
@@ -39,45 +39,44 @@ namespace MeltPoolDG::Heat
...
-                           LaserData<double>::GusarovData gusarov_data_in,
-                           unsigned int                   temp_dof_idx_in);
+    LaserHeatSourceGusarov(const LaserData<double>::GusarovData &gusarov_data_in);
 
     void
-    compute_volumetric_heat_source(VectorType &      heat_source_vector,
-                                   double            laser_power,
-                                   const Point<dim> &laser_position,
-                                   bool              zero_out = true);
+    compute_volumetric_heat_source(VectorType &            heat_source_vector,
```
mschreter, line 53:
```suggestion
/**
* Compute a DoF vector of the heat source from the Gusarov laser model. 
*/
    compute_volumetric_heat_source(VectorType &            heat_source_vector,
```

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
```diff
@@ -0,0 +1,145 @@
...
+// MeltPoolDG
+#include <meltpooldg/heat/laser_heat_source_gusarov.hpp>
+#include <meltpooldg/interface/simulationbase.hpp>
+#include <meltpooldg/utilities/distance_functions.hpp>
+
+/**
+ * This simulation represents simple test examples for heat transfer with melt front propagation.
+ * The problem is inspired by the Proell et al. [1] single track scan example.
+ *
+ * The slap has properties of Ti-6Al-4V and starts of below the solidus temperature and is subjected
```
mschreter, line 19:
```suggestion
 * The slab has properties of Ti-6Al-4V, is initially below the solidus temperature and is subjected
```

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      this->set_parameters();
+    }
+
+    void
+    create_spatial_discretization() override
+    {
+      if constexpr (dim == 1)
+        {
+          this->triangulation = std::make_shared<Triangulation<1>>();
```
mschreter, line 81: Just as a comment: for 1D you could use a `parallel::shared::Triangulation`, as also implemented e.g. in 

https://github.com/MeltPoolDG/MeltPoolDG/blob/372773d844004c4a462de94cc2242519d14b730a/include/meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp#L155-L162

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
```diff
@@ -0,0 +1,145 @@
...
+      const types::boundary_id left_bc = 10;
+
+      for (const auto &cell : this->triangulation->cell_iterators())
+        {
+          for (auto &face : cell->face_iterators())
+            if (face->at_boundary())
+              {
+                if (face->center()[0] == x_max)
+                  face->set_boundary_id(left_bc);
+              }
```
mschreter, line 129: This might be extended when we compute it in 3D, which would be a nice test as well :-).

source/heat/heat_transfer_problem.cpp
```diff
@@ -29,10 +29,15 @@ namespace MeltPoolDG::Heat
         scratch_data->get_pcout() << "t= " << std::setw(10) << std::left
                                   << time_iterator.get_current_time();
 
-        if (base_in->parameters.heat.velocity != 0.0)
-          compute_field_vector(velocity,
-                               velocity_dof_idx,
-                               *base_in->get_velocity_field("heat_transfer"));
+        const auto source_field_function = base_in->get_source_field("heat_transfer");
+        if (source_field_function)
```
mschreter, line 33: I am not entirely sure but couldn't you simply write
```suggestion
        if (const auto source_field_function = base_in->get_source_field("heat_transfer"))
```

source/heat/laser_heat_source_gusarov.cpp
```diff
@@ -63,7 +76,7 @@ namespace MeltPoolDG::Heat
 
   template <int dim>
   double
-  LaserHeatSourceGusarov<dim>::dq_dxi(const double xi)
+  LaserHeatSourceGusarov<dim>::dq_dxi(const double xi) const
   {
     const double &rho = gusarov_data.reflectivity;
     const double  a   = std::sqrt(1 - rho);
```
mschreter, line 82: Maybe we can make `a` a permanent class member.

include/meltpooldg/heat/laser_heat_source_gusarov.hpp
nmuch: Sure. But why not `compute_local_volumetric_heat_source()` instead. Imo having compute in front makes this more readable.

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_gusarov.json
```diff
@@ -0,0 +1,46 @@
...
+  "laser": {
+    "laser power": "30.",
+    "laser center": "0.0,0.0",
+    "laser do move": "false",
+    "laser heat source model": "Gusarov",
+    "laser gusarov laser beam radius": "0.06e-3",
+    "laser gusarov reflectivity": "0.7",
+    "laser gusarov extinction coefficient": "60.0e3",
+    "laser gusarov layer thickness": "0.05e-3"
+  },
```
mschreter, line 30: Maybe you could also investigate if it works with AMR.

mschreter (approved):
Thanks @nmuch, I've only pointed out some minor potential improvements.

source/heat/laser_heat_source_gusarov.cpp
nmuch: Yes, also `D` just below.

nmuch:
> I tried enabling AMR, but sadly it does not work atm.

I found the bug; in the heat operator's tangent methods the ghost values of the old temperature vector are not updated. In the case of solidification the tangent cell operation needs the old temperature values. This is a bug I introduced in #162 and I'm surprised it only came up now. 
Do you want me to add the bugfix to this PR or open another?

mschreter:
> > I tried enabling AMR, but sadly it does not work atm.
> 
> I found the bug; in the heat operator's tangent methods the ghost values of the old temperature vector are not updated. In the case of solidification the tangent cell operation needs the old temperature values. This is a bug I introduced in #162 and I'm surprised it only came up now.
> Do you want me to add the bugfix to this PR or open another?

Let's add it in a follow-up PR, so we can clearly distinguish when (in the unexpected case) the test output changes.

nmuch:
> Let's add it in a follow-up PR

I created #203 just for the bugfix. Then both this PR and #203 are merged I can create a melt front propagation simulation with AMR. Locally I already ran it and the result looks promising :)

Can we proceed here?

mschreter:
Sounds good. Although AMR of the heat problem is not highest priority but should hopefully work out of the box ;-). Thanks.

PR #203, opened 2021-05-26
Bugfix: Update ghost values of temperature_old in tangent operations of heat operator

nmuch:
This is a bug introduced in #162 and came up in #201.


peterrum:
@nmuch Let me ask you how did you find this problem and/or why was this not issue before. Is there no test for parallel with solidification? Was somewhere an assert thrown?

nmuch:
@peterrum When running a simulation with solidification and AMR in parallel, the debug build threw an assert that prompted to update ghost values. The assert was thrown by `FEEvaluation::gather_evaluate()` in the heat tangent operator. Without AMR the simulation works fine. I can only assume the ghost values were still updated in that case. There are tests for heat with solidification run in parallel e.g. `unidirectional_heat_transfer_with_velocity_and_solidification` but none with AMR.

peterrum:
OK. Thanks for the explanation! Everything should be than fine from deal.II side, just wanted to check!

PR #204, opened 2021-05-27
Add simulation and test for melt front propagation with AMR

nmuch:
Since I already worked on this we can benefit from the test case. #201, #203
animation_amr

### DIFF
--- a/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_gusarov_amr.json
+++ b/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_gusarov_amr.json
@@ -1,0 +1,54 @@
+{
+  "base": {
+    "application name": "melt_front_propagation",
+    "degree": "1",
+    "dimension": "2",
+    "do print parameters": "false",
+    "global refinements": "4",
+    "problem name": "heat_transfer",
+    "verbosity level": "2"
+  },
+  "heat": {
+    "heat start time": "0.0",
+    "heat time step size": "5e-6",
+    "heat end time": "0.015",
+    "heat max n steps": "3000",
+    "heat solidification": "true",
+    "heat solver rel tolerance": "1e-12",
+    "heat solver preconditioner type": "ILU",
+    "heat nlsolve residual tolerance": "1e-8"
+  },
+  "laser": {
+    "laser power": "30.",
+    "laser center": "0.0,0.0",
+    "laser do move": "false",
+    "laser heat source model": "Gusarov",
+    "laser gusarov laser beam radius": "0.06e-3",
+    "laser gusarov reflectivity": "0.7",
+    "laser gusarov extinction coefficient": "60.0e3",
+    "laser gusarov layer thickness": "0.05e-3"
+  },
+  "material": {
+    "material second conductivity": "55.563",
+    "material second capacity": "460.0",
+    "material second density": "7850.0",
+    "material solid conductivity": "17.0",
+    "material solid capacity": "700.0",
+    "material solid density": "7850.0",
+    "material solidus temperature": "1966.0",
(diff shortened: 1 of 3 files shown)


peterrum:
The criterion for refinement/coarsening does not seem to make sense. At the end practically everything is uniformly refined. Maybe in a follow-up PR you could take a look at a better criterion for solidification (the first version was targeting the interface between liquid and vapor).

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_gusarov_amr.json
```diff
@@ -0,0 +1,54 @@
...
+    "material solid density": "7850.0",
+    "material solidus temperature": "1966.0",
+    "material liquidus temperature": "1974.0"
+  },
+  "adaptive meshing": {
+    "do amr": "true",
+    "do not modify boundary cells": "false",
+    "upper perc to refine": " 0.2",
+    "lower perc to coarsen": "0.1",
+    "max grid refinement level": "5",
```
mschreter, line 46: At the moment you only perform AMR with a gain of 1 refinement level. Did you also try out a higher value for the 'max grid refinement level' of e.g. 7? What I was expected to see in the results was that we obtain refinement solely in a rather narrow band around the mushy zone. Maybe you could rerun the simulation once more by choosing a lower percentage of refine/coarsen and a higher value for the 'max grid refinement level', and update the animation for this use case. Thanks!

mschreter (commented):
Thanks @nmuch, this test is a good indicator that the `heat_transfer_operation` can handle AMR properly. The latter is important regarding our goal of the melt pool simulations with AMR. I pointed out a minor comment/request. After that I am happy to merge :-).

mschreter:
I just read your comment @peterrum. At the moment refinement is done essentially based on the temperature gradient, which also explains why the refinement starts to evolve from the laser spot. In case of solidification we could think about filling a DoF-Vector `solid_fraction` (or vice-versa `liquid_fraction`) and evaluating the `KellyErrorEstimator` upon the latter. Then in case of solidification we could modify
```cpp
if (solidification)
  locally_relevant_solution.copy_locally_owned_data_from(heat_operation->get_solid_fraction()); 
```
in the `heat_transfer_problem`.
https://github.com/MeltPoolDG/MeltPoolDG/blob/da17658153b6ea8bd9fc935d97cae083948d81a3/source/heat/heat_transfer_problem.cpp#L323-L334

 @nmuch: What do you think?

nmuch:
> At the moment refinement is done essentially based on the temperature gradient, which also explains why the refinement starts to evolve from the laser spot.

I was about to comment the same. The error estimator is not informed by the solid/liquid fraction atm. Your suggestion to use the solid fraction as `locally_relevant_solution` sound good. However, imo the error estimator should also consider steep temperature gradients regardless of the solid fraction. Is it possible to combine the two fields for the error estimator? I can look into this for a follow-up.

mschreter:
We could think about evaluating the `KellyErrorEstimator` twice: one evaluation of the `temperature` (resulting in `estimated_temperature_error_per_cell`) and one of the `solid_fraction` (resulting in `estimated_solidification_error_per_cell`). Then we merge the two vectors `estimated_temperature_error_per_cell` and `estimated_solidification_error_per_cell` into one `estimated_error_per_cell` using a certain criteria. I am not sure if `max(estimated_solidification_error_per_cell, estimated_temperature_error_per_cell)` is sufficient. Maybe you can investigate that in a follow-up PR.

nmuch:
Thank you @mschreter for your review. I made some changes to the AMR parameters as you pointed out. Here's what the simulation looks now. I'm now showing the temperature field instead of the liquid fraction because it's the relevant field for AMR.
animation_amr_temperature

Issue #213, opened 2021-06-01
Refactor Evaporation Operation


mschreter:
The overall goal is to split the tasks

- evaporation model --> determines mDot(temperature)
- compute mDot DoF-Vector --> 3 variants
- compute source terms (level set, mass balance, heat) --> 2 variants sharp/continuous

PR #215, opened 2021-06-01
Refactor evaporative mass flux computation

mschreter:
This PR splits up the variants to fill the DoF vector for the evaporative mass flux:

- continuous over interface (`evaporation_mass_flux_operator_continuous.hpp`) --> take the value of mDot as it is
- interface value (`evaporation_mass_flux_operator_interface_value.hpp`) --> take the value of mDot at the interface and spread it across the interface

references #213


peterrum (approved):
Looks good. I have directly pushed some minor modifications.

mschreter:
Thanks a lot!

PR #219, opened 2021-06-02
Set matrix free default value to true and remove deleted parameter "do print l2norm"

### LINKED ISSUE #217: Set default value for "matrix free" to true?
What do you think @peterrum / @nmuch?

mschreter:
closes #217


nmuch (approved):
Thanks!

PR #220, opened 2021-06-02
Set "do reinitialization" default value to true

mschreter:
... this has no effect on any of the test cases.


mschreter:
I'll rebase once #219 is merged.

PR #223, opened 2021-06-03
Enable DEAL_II_DEPRECATED_EARLY per default

peterrum:
~... depends on https://github.com/dealii/dealii/pull/12383.~

The reason for making this change is to be able to handle depredations as early as possible, however, the official Docker image is compiled without `DEAL_II_DEPRECATED_EARLY` enabled so that one needs to shift the responsibility to the user code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,12 @@ add_library(${libName} ${sources})
 
 deal_ii_setup_target(${libName})
 
+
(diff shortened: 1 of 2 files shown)


peterrum:
OK. This worked with the newest deal.II master docker image :) Compilation failed as expected! Let's merge once we are not compatible with deal.II 9.3 anymore.

mschreter (approved):

PR #224, opened 2021-06-03
Fix DEAL_II_DEPRECATED_EARLY stuff

peterrum:
"Noticed" while working on #223. I actually expected that these will give us trouble once `DEAL_II_DEPRECATED_EARLY` have been converted to `DEAL_II_DEPRECATED` directly after the deal.II release. But this did not happen, since we have forgot to add the `DEAL_II_DEPRECATED_EARLY` label. It has been fixed in https://github.com/dealii/dealii/pull/12350 so that the bools are available one more year...


mschreter:
Thanks!

mschreter (approved):

PR #229, opened 2021-06-04
Fix UtilityFunctions::fill_dof_vector_from_cell_operation_vec() for higher degrees

peterrum:

PR #230, opened 2021-06-05
Introduce a single fill_dof_vector_from_cell_operation()

peterrum:
I just wanted to keep this out of PR #229. ~Only the last commit is relevant.~

Pre C++17, one could accomplish the same with specialized template functions. In most cases, the language does not actually adds anything new but "improves" the syntax.


mschreter:
Thanks Peter. Since I (we) prefer to use "modern" C++, I don't think having C++17(20) features should be an issue.

mschreter (approved):

PR #232, opened 2021-06-07
Fix melt front propagation laser direction

nmuch:
This PR fixes the domain of the melt front simulation, so that the laser heat soure acts opon the material correctly.
Fyi @mschreter

### DIFF
--- a/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
+++ b/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
@@ -103,8 +103,8 @@ namespace MeltPoolDG::Simulation::MeltFrontPropagation
           std::vector<unsigned> refinements(2, 1);
           refinements[0] = 3;
           // create mesh
-          const Point<2> left(0, 0);
-          const Point<2> right(x_max, y_max);
+          const Point<2> left(0, -z_max);
+          const Point<2> right(x_max, 0);
           GridGenerator::subdivided_hyper_rectangle(*this->triangulation, refinements, left, right);
           this->triangulation->refine_global(this->parameters.base.global_refinements);
         }
@@ -117,7 +117,7 @@ namespace MeltPoolDG::Simulation::MeltFrontPropagation
           refinements[1] = 3;
           // create mesh
           const Point<3> left(0, 0, 0);
-          const Point<3> right(x_max, y_max, z_max);
+          const Point<3> right(x_max, y_max, -z_max);
           GridGenerator::subdivided_hyper_rectangle(*this->triangulation, refinements, left, right);
           this->triangulation->refine_global(this->parameters.base.global_refinements);
         }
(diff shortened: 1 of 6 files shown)


mschreter:
Does this change have to do with the following line, indicating that the heated volume has to be below the applied laser heat source? If yes, I would add here a comment/assert. 

https://github.com/MeltPoolDG/MeltPoolDG/blob/32cbc019ffa0dcb2c195cbc68f3ebd65d2002473/source/heat/laser_heat_source_gusarov.cpp#L23

nmuch:
> Does this change have to do with the following line, indicating that the heated volume has to be below the applied laser heat source? 

Correct. Otherwise, the heat source model does not make mush sense.

>If yes, I would add here a comment/assert.
> 
> https://github.com/MeltPoolDG/MeltPoolDG/blob/32cbc019ffa0dcb2c195cbc68f3ebd65d2002473/source/heat/laser_heat_source_gusarov.cpp#L23

I think an assert does not really work here, because this equation is evaluated for the entire domain which also includes the surrounding "air"-domain. 
I think the doc describes the usage of this laser model quite well
https://github.com/MeltPoolDG/MeltPoolDG/blob/32cbc019ffa0dcb2c195cbc68f3ebd65d2002473/include/meltpooldg/heat/laser_heat_source_gusarov.hpp#L19-L36
We can think about adding a coordinate system to this sketch in addition to the comment https://github.com/MeltPoolDG/MeltPoolDG/blob/32cbc019ffa0dcb2c195cbc68f3ebd65d2002473/include/meltpooldg/heat/laser_heat_source_gusarov.hpp#L54-L55

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
```diff
@@ -104,7 +104,7 @@ namespace MeltPoolDG::Simulation::MeltFrontPropagation
           refinements[0] = 3;
           // create mesh
           const Point<2> left(0, 0);
-          const Point<2> right(x_max, y_max);
+          const Point<2> right(x_max, -z_max);
```
peterrum, line 107:
```suggestion
          const Point<2> left(0, -z_max);
          const Point<2> right(x_max, 0);
```
I guess deal.II will re-orient the cells internally, but the notion is normally `left-bottom-front` and `right-top-back`.

mschreter:
> We can think about adding a coordinate system to this sketch in addition to the comment
> https://github.com/MeltPoolDG/MeltPoolDG/blob/32cbc019ffa0dcb2c195cbc68f3ebd65d2002473/include/meltpooldg/heat/laser_heat_source_gusarov.hpp#L54-L55

This would be a good idea!

mschreter (approved):

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation.hpp
mschreter: @peterrum: Do you see an issue here?

nmuch: I already applied @peterrum 's suggestion

PR #233, opened 2021-06-07
Remove level_set_with_evaporation from level set problem

mschreter:
This removes the usage of `LevelSetWithEvaporationOperation` within the `LevelSetProblem`. As a follow-up, the same will be done for the `TwoPhaseFlowProblem`.

depends on #241.


tests/rotating_bubble_with_evaporation.mpirun=4.output
```diff
@@ -8,28 +8,31 @@
  	 |n_0| = 4.6693707854     |n_1| = 4.6693716486    
 |k| = 4.2995266374    
 | ls: t= 0.01       	 |n_0| = 4.6693707854     |n_1| = 4.6693716486    
+|vel|= 424.59165838
```
mschreter, line 11: This is the major change in the test output --> the computed velocity is now also printed (stemming from the `advection_operation->solve()` function).

tests/rotating_bubble_with_evaporation.mpirun=4.output
```diff
@@ -1,4 +1,4 @@
- 	 |n_0| = 4.9221903419     |n_1| = 4.9221910318    
+ 	 |n_0| = 4.9221905105     |n_1| = 4.9221910318
```
mschreter, line 1: The numbers slightly changed since I produced the output locally in release configuration.

peterrum (approved):
Looking forward to the follow-up PR. Then we can eliminate a couple hundred locs :)

peterrum:
This is odd! The error in the case of `tests/stefans_problem_mf.mpirun=4.release` is only for the velocity!? Do you need to call update ghost values/distribute constraints? The error in the other case is very small.

mschreter:
> This is odd! The error in the case of `tests/stefans_problem_mf.mpirun=4.release` is only for the velocity!? Do you need to call update ghost values/distribute constraints? The error in the other case is very small.

I am currently not sure where the difference comes from. Ghost values of the velocity should be up-to-date. However, I observed that the time steps in `rotating_bubble_with_evaporation` were rather big and the level set was not resolved properly. I will have a look at the `stefans_problem_mf` as well.

mschreter:
Mmh. These are the two lines where the failing output is produced:

mschreter:
Obviously also the serial test also fails...

tests/rotating_bubble_with_evaporation.json
```diff
@@ -47,4 +47,4 @@
   "paraview": {
     "paraview do output": "false"
   }
-}
\ No newline at end of file
+}
```
peterrum: Ping!

peterrum:
Could you rebase this PR!

PR #235, opened 2021-06-07
Add interface laser heat source

nmuch:
This PR adds an alternative laser heat source impact model that only acts on the level-set interface. 

For the Gauss laser model this is implemented according to the SPH paper. 
The interface laser model is not compatible with the Gusarov laser model, so there is an assent. 

The laser heat source of any kind is now applied to the heat_source dof-vector by `HeatTransferProblem::run()`. 

A case and test has been added to the melt front propagation simulation, that includes a prescribed level-set. The domain has been expanded by a gaseous domain in positive dim-1 direction.

~~This PR includes #232. I'll rebase after #232 is merged.~~

### DIFF
--- a/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
+++ b/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
@@ -61,21 +61,21 @@ namespace MeltPoolDG::Heat
                                          double            power) const final;
 
     void
-    compute_interface_heat_source(VectorType &            heat_source_vector,
-                                  const ScratchData<dim> &scratch_data,
-                                  unsigned int            temp_dof_idx,
-                                  double                  laser_power,
-                                  const Point<dim> &      laser_position,
-                                  const VectorType &      level_set_heaviside,
-                                  unsigned int            ls_dof_idx,
-                                  bool                    zero_out = true) const final;
+    compute_interfacial_heat_source(VectorType &            heat_source_vector,
+                                    const ScratchData<dim> &scratch_data,
+                                    unsigned int            temp_dof_idx,
+                                    double                  laser_power,
+                                    const Point<dim> &      laser_position,
+                                    const VectorType &      level_set_heaviside,
+                                    unsigned int            ls_dof_idx,
+                                    bool                    zero_out = true) const final;
 
     double
-    local_compute_interface_heat_source(const Point<dim> &            position,
-                                        const Point<dim> &            laser_position,
-                                        double                        power,
-                                        const Tensor<1, dim, double> &normal_vector,
-                                        double                        delta_value) const final;
+    local_compute_interfacial_heat_source(const Point<dim> &            position,
+                                          const Point<dim> &            laser_position,
+                                          double                        power,
+                                          const Tensor<1, dim, double> &normal_vector,
+                                          double                        delta_value) const final;
 
   private:
     /**
--- a/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
+++ b/include/meltpooldg/heat/laser_heat_source_gusarov.hpp
@@ -60,6 +60,23 @@ namespace MeltPoolDG::Heat
                                          const Point<dim> &laser_position,
                                          double            power) const final;
 
+    void
+    compute_interface_heat_source(VectorType &            heat_source_vector,
+                                  const ScratchData<dim> &scratch_data,
+                                  unsigned int            temp_dof_idx,
+                                  double                  laser_power,
+                                  const Point<dim> &      laser_position,
+                                  const VectorType &      level_set_heaviside,
+                                  unsigned int            ls_dof_idx,
+                                  bool                    zero_out = true) const final;
+
+    double
+    local_compute_interface_heat_source(const Point<dim> &            position,
+                                        const Point<dim> &            laser_position,
+                                        double                        power,
+                                        const Tensor<1, dim, double> &normal_vector,
+                                        double                        delta_value) const final;
+
   private:
     /**
      * Equation (26) corrected to match Figure 5 of Gusarov et al. (2009)
--- a/include/meltpooldg/interface/parameters.hpp
+++ b/include/meltpooldg/interface/parameters.hpp
@@ -172,7 +172,7 @@ namespace MeltPoolDG
     bool        do_move                            = false;
     number      scan_speed                         = 0.0;
     bool        variable_properties_over_interface = false;
-    std::string impact                             = "volumetric";
+    std::string impact_type                        = "volumetric";
     std::string heat_source_model                  = "Gusarov";
     struct GusarovData
     {
--- a/include/meltpooldg/interface/parameters.hpp
+++ b/include/meltpooldg/interface/parameters.hpp
@@ -172,6 +172,7 @@ namespace MeltPoolDG
     bool        do_move                            = false;
     number      scan_speed                         = 0.0;
     bool        variable_properties_over_interface = false;
+    std::string impact                             = "volumetric";
     std::string heat_source_model                  = "Gusarov";
     struct GusarovData
     {
--- a/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_interface_gauss.json
+++ b/include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_interface_gauss.json
@@ -1,0 +1,50 @@
+{
+  "base": {
+    "application name": "melt_front_propagation",
+    "degree": "1",
+    "dimension": "2",
+    "do print parameters": "false",
+    "global refinements": "5",
+    "problem name": "heat_transfer",
+    "verbosity level": "2"
+  },
+  "heat": {
+    "heat start time": "0.0",
+    "heat time step size": "1e-4",
+    "heat end time": "0.015",
+    "heat max n steps": "150",
+    "heat velocity": "-1",
+    "heat solidification": "true",
+    "heat two phase": "true",
+    "heat solver rel tolerance": "1e-12",
+    "heat solver preconditioner type": "ILU",
+    "heat nlsolve residual tolerance": "1e-8"
+  },
+  "laser": {
+    "laser power": "10.",
+    "laser center": "0.0,0.0",
+    "laser do move": "false",
+    "laser heat source model": "Gauss",
+    "laser impact": "interface",
+    "laser gauss laser beam radius": "0.06e-3",
+    "laser gauss absorptivity": "0.5"
+  },
+  "material": {
+    "material first capacity": "0",
+    "material first conductivity": "10",
+    "material first density": "74.30",
+    "material second conductivity": "55.563",
+    "material second capacity": "460.0",
+    "material second density": "7850.0",
+    "material solid conductivity": "17.0",
+    "material solid capacity": "700.0",
+    "material solid density": "7850.0",
+    "material solidus temperature": "1966.0",
+    "material liquidus temperature": "1974.0"
+  },
+  "paraview": {
+    "paraview do output": "true",
+    "paraview directory": "output",
+    "paraview filename": "melt_front_propagation_static_surface_gauss"
+  }
+}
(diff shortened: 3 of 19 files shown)


source/heat/heat_transfer_problem.cpp
```diff
@@ -42,6 +39,38 @@ namespace MeltPoolDG::Heat
                                level_set_dof_idx,
                                *base_in->get_initial_condition("prescribed_level_set"));
 
+        if (const auto source_field_function = base_in->get_source_field("heat_transfer"))
+          compute_field_vector(heat_operation->get_heat_source(),
+                               temp_dof_idx,
+                               *source_field_function);
+        // TODO: Atm the laser heat source will overwrite the heat source field function if both are
+        // given. Instead they should be added.
```
mschreter, line 47: This could be easily integrated by replacing `=` by `+=` in 
https://github.com/MeltPoolDG/MeltPoolDG/blob/53cf72d93566636b4b2a0ef33d67374d3b3d763a/source/heat/laser_heat_source_base.cpp#L38-L42

and setting `zero_out` to false in the subsequent lines.

include/meltpooldg/heat/laser_heat_source_gusarov.hpp
```diff
@@ -60,6 +60,23 @@ namespace MeltPoolDG::Heat
                                          const Point<dim> &laser_position,
                                          double            power) const final;
 
+    void
+    compute_interface_heat_source(VectorType &            heat_source_vector,
```
mschreter, line 64:
```suggestion
    compute_interfacial_heat_source(VectorType &            heat_source_vector,
```

include/meltpooldg/heat/laser_heat_source_gusarov.hpp
```diff
@@ -60,6 +60,23 @@ namespace MeltPoolDG::Heat
...
+                                  const ScratchData<dim> &scratch_data,
+                                  unsigned int            temp_dof_idx,
+                                  double                  laser_power,
+                                  const Point<dim> &      laser_position,
+                                  const VectorType &      level_set_heaviside,
+                                  unsigned int            ls_dof_idx,
+                                  bool                    zero_out = true) const final;
+
+    double
+    local_compute_interface_heat_source(const Point<dim> &            position,
```
mschreter, line 74:
```suggestion
    local_compute_interfacial_heat_source(const Point<dim> &            position,
```

include/meltpooldg/interface/parameters.hpp
```diff
@@ -172,6 +172,7 @@ namespace MeltPoolDG
     bool        do_move                            = false;
     number      scan_speed                         = 0.0;
     bool        variable_properties_over_interface = false;
+    std::string impact                             = "volumetric";
```
mschreter, line 175:
```suggestion
    std::string impact_type                             = "volumetric";
```

include/meltpooldg/simulations/melt_front_propagation/melt_front_propagation_static_interface_gauss.json
```diff
@@ -0,0 +1,50 @@
...
+    "heat solver rel tolerance": "1e-12",
+    "heat solver preconditioner type": "ILU",
+    "heat nlsolve residual tolerance": "1e-8"
+  },
+  "laser": {
+    "laser power": "10.",
+    "laser center": "0.0,0.0",
+    "laser do move": "false",
+    "laser heat source model": "Gauss",
+    "laser impact": "interface",
```
mschreter, line 28:
```suggestion
    "laser impact type": "interface",
```

source/heat/laser.cpp
```diff
@@ -19,8 +19,8 @@ namespace MeltPoolDG::Heat
   {
     current_time = start_time;
     compute_laser_intensity();
-    scratch_data.get_pcout() << "current laser position: " << laser_position << std::endl;
-    scratch_data.get_pcout() << "current laser intensity: " << laser_intensity << std::endl;
+    scratch_data.get_pcout(1) << "current laser position: " << laser_position << std::endl;
+    scratch_data.get_pcout(1) << "current laser intensity: " << laser_intensity << std::endl;
```
mschreter, line 23: In my opinion, it makes sense to have the laser position/intensity in the test output, in particular if we have a moving laser heat source.

source/heat/laser_heat_source_gauss.cpp
```diff
@@ -20,15 +22,118 @@ namespace MeltPoolDG::Heat
...
-  LaserHeatSourceGauss<dim>::power_density(const double radius, const double power) const
+  LaserHeatSourceGauss<dim>::local_compute_interface_heat_source(
+    const Point<dim> &            position,
+    const Point<dim> &            laser_position,
+    double                        power,
+    const Tensor<1, dim, double> &normal_vector,
+    double                        delta_value) const
+  {
+    // only consider distance in x (2D) or x and y (3D) direction. Disregard distance in dim-1
+    // direction.
```
mschreter, line 38: Could you please explain, why this modification is necessary?

source/heat/laser_heat_source_gauss.cpp
```diff
@@ -20,15 +22,118 @@ namespace MeltPoolDG::Heat
...
+      scratch_data.get_dof_handler(temp_dof_idx).get_fe(),
+      Quadrature<dim>(
+        scratch_data.get_dof_handler(temp_dof_idx).get_fe().get_unit_support_points()),
+      update_quadrature_points);
+
+    level_set_heaviside.update_ghost_values();
+    FEValues<dim> ls_heaviside_eval(
+      scratch_data.get_mapping(),
+      scratch_data.get_dof_handler(ls_dof_idx).get_fe(),
+      Quadrature<dim>(scratch_data.get_dof_handler(ls_dof_idx).get_fe().get_unit_support_points()),
```
mschreter, line 79:
```suggestion
      Quadrature<dim>(scratch_data.get_dof_handler(temp_dof_idx).get_fe().get_unit_support_points()),
```

source/interface/parameters.cpp
```diff
@@ -601,6 +601,11 @@ namespace MeltPoolDG
       prm.add_parameter("laser gusarov layer thickness",
                         laser.gusarov.layer_thickness,
                         "Layer thickness");
+      prm.add_parameter(
+        "laser impact",
+        laser.impact,
```
mschreter, line 606:
```suggestion
        "laser impact type",
        laser.impact_type,
```

source/heat/heat_transfer_operation.cpp
```diff
@@ -260,6 +260,13 @@ namespace MeltPoolDG::Heat
     return heat_source;
   }
 
+  template <int dim>
+  const VectorType &
+  HeatTransferOperation<dim>::get_level_set_as_heaviside() const
+  {
+    return *level_set_as_heaviside;
```
mschreter, line 267:
```suggestion
    Assert(level_set_as_heaviside, ExcMessage("You requested the level set vector from the heat operation, which is a nullptr. You must provide a valid level_set_as_heaviside pointer to the heat operation."));
    return *level_set_as_heaviside;
```

source/heat/heat_transfer_problem.cpp
```diff
@@ -42,6 +39,38 @@ namespace MeltPoolDG::Heat
...
+            else if (base_in->parameters.laser.impact == "interface")
+              laser_heat_source_operation->compute_interface_heat_source(
+                heat_operation->get_heat_source(),
+                *scratch_data,
+                temp_dof_idx,
+                laser_operation->get_laser_power(),
+                laser_operation->get_laser_position(),
+                heat_operation->get_level_set_as_heaviside(),
+                level_set_dof_idx,
+                true /* zero_out */);
```
mschreter, line 69: We could think about moving it into a `compute_heat_source()` operation within the `LaserHeatSourceBase` class -- but not now :-).

source/heat/heat_transfer_problem.cpp
```diff
@@ -156,6 +185,33 @@ namespace MeltPoolDG::Heat
...
+        else if (base_in->parameters.laser.heat_source_model == "Gauss")
+          {
+            laser_heat_source_operation =
+              std::make_shared<Heat::LaserHeatSourceGauss<dim>>(base_in->parameters.laser.gauss);
+          }
+        else
+          AssertThrow(false,
+                      ExcMessage(
+                        "No requested laser model found. Please speficy the "
+                        "heat source model in the laser section of the input parameters."));
```
mschreter, line 212: We could think about moving it into the `LaserHeatSourceBase` class -- but not now :-).

source/heat/laser_heat_source_gauss.cpp
```diff
@@ -20,15 +22,118 @@ namespace MeltPoolDG::Heat
...
+
+            for (const auto q : heat_source_eval.quadrature_point_indices())
+              {
+                const double delta_value = grad_ls_heaviside[q].norm();
+                if (delta_value == 0.0)
+                  {
+                    heat_source_vector[local_dof_indices[q]] = 0.0;
+                    continue;
+                  }
+                const Tensor<1, dim, double> normal_vector = grad_ls_heaviside[q] / delta_value;
```
mschreter, line 108: Here, we might face problems since we are not using the diffused `normal_vector`.

```suggestion
// TODO: use computed normal vector
                const Tensor<1, dim, double> normal_vector = grad_ls_heaviside[q] / delta_value;
```

mschreter (approved):
Thanks @nmuch. I have some minor suggestions. As discussed, the "cleaner" way without the restriction of `ls_degree=temp_degree` would be an algorithm similar to the following:

https://github.com/MeltPoolDG/MeltPoolDG/blob/d123b475db00c42e46aa21dfd9b5d794bb11478d/source/evaporation/evaporation_source_terms_continuous.cpp#L46-L152

source/heat/laser.cpp
mschreter: This change is the reason why the tests fail. Could you please revert it?

source/heat/laser.cpp
```diff
@@ -35,8 +35,8 @@ namespace MeltPoolDG::Heat
     // 2) update intensity of the laser
     compute_laser_intensity();
 
-    scratch_data.get_pcout() << "current laser position: " << laser_position << std::endl;
-    scratch_data.get_pcout() << "current laser intensity: " << laser_intensity << std::endl;
+    scratch_data.get_pcout(1) << "current laser position: " << laser_position << std::endl;
+    scratch_data.get_pcout(1) << "current laser intensity: " << laser_intensity << std::endl;
```
mschreter, line 39: Same here.

source/heat/heat_transfer_problem.cpp
nmuch: Yes that looks like an easy fix but the cell operation in `local_compute_volumetric_heat_source()` touches every dof of every cell. This means that most dof entries are being written multiple times. The `+=` operator would result in a wrong heat source vector. 
I want to find a solution to this problem in a followup.

source/heat/laser.cpp
nmuch: Yes. Then I have to adapt some melt front test outputs..

source/heat/laser_heat_source_gauss.cpp
nmuch: The laser position is a point, but the laser operates along the entire `dim-1`-axis. This modification gets rid of any distance in `dim-1`-direction. I imagine the laser as a cylinder here. The heat source is then computated in the intersection of that cylinder and der level-set interface.

nmuch:
Thank you @mschreter for your great review! I applied the quick suggestions. I'll look into https://github.com/MeltPoolDG/MeltPoolDG/pull/235#pullrequestreview-677526612 tomorrow.

source/heat/heat_transfer_problem.cpp
mschreter: I see. It might become an issue when I fill the `heat_source` from the `evaporation_operation`. Let's keep this in mind and postpone it.

PR #240, opened 2021-06-08
Evaporative mass flux from interface value: bug fix + test file

mschreter:
This PR fixes a bug in the computation of the evaporative mass flux from the interface value:

```math
\dot{m}(x) = \dot{m}(x_i) =  \dot{m}(x-d\cdot\boldsymbol{n}) 
```

and adds a test file.

Further, a utility function to compute the mesh-size-dependent numerical zero is introduced. The latter will be integrated into the remaining code in a follow-up PR.


source/evaporation/evaporation_mass_flux_operator_interface_value.cpp
```diff
@@ -112,32 +124,35 @@ namespace MeltPoolDG::Evaporation
...
     evaporative_mass_flux = 0.0;
 
-    for (unsigned int i = 0; i < evaluation_points.size(); i++)
+    for (unsigned int i = 0; i < evaluation_points.size(); ++i)
       {
-        evaporative_mass_flux[temp_local_dof_indices[i]] =
+        evaporative_mass_flux[dof_indices[i]] =
           evaporation_model.local_compute_evaporative_mass_flux(
-            temperature[temp_local_dof_indices[i]]);
+            temperature_evaluation_values[dof_indices[i]]);
```
peterrum, line 155:
```suggestion
            temperature_evaluation_values[i]);
```
Shouldn't this be like this? `temperature_evaluation_values` is returned by PRE.

peterrum (commented):
Looks good. One question.

source/evaporation/evaporation_mass_flux_operator_interface_value.cpp
mschreter: Eagle eyes, true! I already integrated into the current status, where I included the number of iterations. Commit follows asap.

peterrum:
Could you rebese. There is a merge conflict.

mschreter:
Using a couple of iterative projects we end up with this. I really like the overlap with adaflo :sunglasses: 

image

peterrum:
> Using a couple of iterative projects we end up with this. I really like the overlap with adaflo sunglasses

I agree!

FYI @JPmoep @kronbichler

PR #242, opened 2021-06-08
Add heat operation solve to melt pool operation

nmuch:
This PR enables solving a heat problem in the melt pool operation with any kind of laser heat source. 

A simple simulation case is provided, that only includes two-phase flow and a stationary interface laser heat problem. It does not include evaporation, recoil pressure, solidification or surface tension. All of these effects should be able to be enabled or will be included soon.


source/heat/heat_transfer_problem.cpp
```diff
@@ -48,16 +48,16 @@ namespace MeltPoolDG::Heat
...
+            if (impact_type == volumetric)
               laser_heat_source_operation->compute_volumetric_heat_source(
                 heat_operation->get_heat_source(),
                 *scratch_data,
                 temp_dof_idx,
                 laser_operation->get_laser_power(),
                 laser_operation->get_laser_position(),
                 true /* zero_out */);
-            else if (base_in->parameters.laser.impact_type == "interface")
+            else if (impact_type == interface)
```
mschreter, line 60: Since you introduced an `enum` you could also use a `switch` statement.

source/melt_pool/melt_pool_operation.cpp
```diff
@@ -147,13 +150,28 @@ namespace MeltPoolDG::MeltPool
...
       {
-        laser_heat_source_operation->compute_volumetric_heat_source(
-          heat_operation->get_heat_source(),
-          *scratch_data,
-          temp_dof_idx,
-          laser_operation->get_laser_power(),
-          laser_operation->get_laser_position(),
-          true /* zero_out */);
+        const auto impact_type = laser_operation->get_laser_impact_type();
+        if (impact_type == Heat::volumetric)
```
mschreter, line 154: same here.

include/meltpooldg/heat/laser.hpp
```diff
@@ -14,6 +14,12 @@ namespace MeltPoolDG::Heat
 {
   using namespace dealii;
 
+  enum LaserImpactType
```
mschreter, line 17: Could you use 
```suggestion
  enum class LaserImpactType
```
to not potentially spoil the namespace?

include/meltpooldg/heat/laser.hpp
```diff
@@ -40,6 +46,8 @@ namespace MeltPoolDG::Heat
      */
     double laser_intensity;
 
+    LaserImpactType impact_type;
```
mschreter, line 49: `const`?

mschreter (approved):
Nice, thanks @nmuch. I have only minor suggestions.

include/meltpooldg/heat/laser.hpp
nmuch: This is not const because of this statement in the ctor:
```
    if (laser_data.impact_type == "volumetric")
      impact_type = volumetric;
    else if (laser_data.impact_type == "interface")
      impact_type = interface;
    else
      AssertThrow(false, ExcMessage("Unknown laser impact type! Abort..."));
```

Do you have a suggestion make this work with `const LaserImpactType impact_type; `?

source/heat/laser.cpp
```diff
@@ -11,7 +11,14 @@ namespace MeltPoolDG::Heat
     , material(material_data_in)
     , laser_position(
         MeltPoolDG::UtilityFunctions::convert_string_coords_to_point<dim>(laser_data.center))
-  {}
+  {
+    if (laser_data.impact_type == "volumetric")
+      impact_type = volumetric;
+    else if (laser_data.impact_type == "interface")
+      impact_type = interface;
```
peterrum: You could move this into the constructor list with `impact_type(to_enum(laser_data.impact_type))` and create a free function for the conversion.

include/meltpooldg/heat/laser.hpp
mschreter: You could try:
```suggestion
    impact_type( laser_data.impact_type == "volumetric" ? volumetric : interface )
```
The assert can be omitted since it is already caught within the parameter reader.

include/meltpooldg/melt_pool/melt_pool_operation.hpp
```diff
@@ -69,17 +69,20 @@ namespace MeltPoolDG
...
+                        VectorType *                                    velocity_in,
+                        unsigned int                                    temp_dof_idx_in,
+                        unsigned int                                    temp_quad_idx_in,
+                        double                                          start_time_in);
 
       void
-      set_initial_condition(const VectorType &level_set_as_heaviside, VectorType &level_set);
+      set_initial_condition(const VectorType &                   level_set_as_heaviside,
+                            VectorType &                         level_set,
+                            std::function<Function<dim> &(void)> get_initial_temperature);
```
peterrum, line 85: I don't get this. Why not use a `std::shared_ptr<Function<dim> >`?

source/heat/laser.cpp
peterrum: If these will be the only options in the future, you could also do it like Magdalena has suggested.

include/meltpooldg/melt_pool/melt_pool_operation.hpp
nmuch: In case the temperature field is determined via `laser_analytical_temperature_field` the initial condition is probably not specified. In that case, calling `base_in->get_initial_condition("heat_transfer")` in the two_phase_flow_problem will lead to a nice segfault. To avoid the segfault I moved the call into a lambda function.

Since the melt_pool_operation does not know the simulation_base I made this quick hack to work around the issue instead of significant restructuring.

include/meltpooldg/heat/laser.hpp
nmuch: I think I'll go with @peterrum 's solution because it can include the Assert if `laser_data.impact_type` is set to a string that does not fit the pattern.

include/meltpooldg/melt_pool/melt_pool_operation.hpp
peterrum: This does not sound right. You can check if a `std::share_ptr` is empty or not.

peterrum:
>The following tests FAILED:
	 13 - tests/melt_pool_tfi_static_interface_gauss_no_solidification.mpirun=4.debug (Failed)

It seems you need to rebase and update the test output.

include/meltpooldg/melt_pool/melt_pool_operation.hpp
nmuch: I pretty sure the problem is here:
https://github.com/MeltPoolDG/MeltPoolDG/blob/6a1a60a28115722d38e86051ceaa742c7b9f6c21/include/meltpooldg/interface/simulation_base.hpp#L300-L304
Where `field_conditions_map` is a std::map. If the field "heat_transfer" has never been set, you have a segfault.

Another solution would be to change these field function getters as follows:
```
    std::shared_ptr<Function<dim>>
    get_initial_condition(const std::string operation_name)
    {
      if (field_conditions_map[operation_name])
        return field_conditions_map[operation_name]->initial_field;
      else
        return nullptr;
    }
```
wdyt?

peterrum: Now, I understand what you do. You create the lambda both in the case `laser_analytical_temperature_field` is enabled and disabled; and the lambda is only executed in the second case (lazy evaluation). I prefer much more the solution you have proposed with the null pointers to this unnecessary indirection. Aren't there any checks in the other getter functions?

mschreter: Could we add an assert here similar to

https://github.com/MeltPoolDG/MeltPoolDG/blob/6a1a60a28115722d38e86051ceaa742c7b9f6c21/source/advection_diffusion/advection_diffusion_problem.cpp#L235-L242

for cases where an initial condition is a must? The IC must be provided for all `melt_pool` type problems if  `laser_heat_source_model != "Analytical"`

```cpp
AssertThrow(*base_in->get_initial_condition("heat_transfer"), ExcMessage("...."));
```

nmuch: Exactly :+1: 
> Aren't there any checks in the other getter functions?

No, there aren't. Let me add them in this PR since the check is needed here.

peterrum: @mschreter The problem is that how the setup works: it that, `get_initial_condition()` is called both for `laser_analytical_temperature_field` enabled and disabled. One if-statement will be needed: either in the simulationbase (checking if the map is empty) or the meltpool-problem class, which just would call `get_initial_condition()` if it is actually needed. Personally, I would prefer the second solution. 

Apart from that, so kind of checks in the getter functions in simulationbase would be also a good idea.

mschreter: That's true. Returning simply a `nullptr` in the default case might also lead to an unexpected behavior. @peterrum I agree and would also suggest to incorporate the `Assert` into the `SimulationBase` class, which might also eliminate a number of code lines (see #249) in our supported problems. If we do not need the definition of an `initial_field_function`, we should actively provide a `nullptr` in e.g. the problem-specific and parameter-dependent case.

nmuch: @mschreter your suggestion sound good to avoid returning nullptr by default. Shall I revert the last commit?
> If we do not need the definition of an initial_field_function, we should actively provide a nullptr in e.g. the problem-specific and parameter-dependent case.

I think the only case that requires the provided nullptr is `recoil_pressure.hpp` then.

source/melt_pool/melt_pool_operation.cpp
```diff
@@ -119,7 +119,8 @@ namespace MeltPoolDG::MeltPool
         laser_operation->get_laser_position());
     else
       {
-        heat_operation->set_initial_condition(get_initial_temperature());
+        if (initial_temperature)
```
mschreter, line 122: You can omit this if statement since in this case the initial condition is a must.

include/meltpooldg/simulations/recoil_pressure/recoil_pressure.hpp
```diff
@@ -190,6 +190,7 @@ namespace MeltPoolDG
           this->attach_initial_condition(std::shared_ptr<Function<dim>>(
                                            new Functions::ZeroFunction<dim>(dim)),
                                          "navier_stokes_u");
+          this->attach_initial_condition(std::shared_ptr<Function<dim>>(), "heat_transfer");
```
mschreter, line 193: Is this really needed?

nmuch: Atm yes.
> If we do not need the definition of an initial_field_function, we should actively provide a nullptr in e.g. the problem-specific and parameter-dependent case.

Maybe rebasing on #250 make this obsolete.

mschreter: I took the liberty to make some minor modifications. In this case #250 would have not fixed it, since we don't want to specify an initial temperature field if the analytical formulation is chosen. Once CI has passed we go ahead and merge. Thanks once more!

PR #244, opened 2021-06-08
Add melt front propagation simulation with moving laser

nmuch:
@mschreter since you keep asking about with feature, I quickly tried it and found no issues 👍 
Here is a simulation and a test case.


mschreter:
Just because I like colorful pictures :-).

animation_reduced

PR #250, opened 2021-06-09
Resolve: Include Asserts for field functions into the simulation_base file

### LINKED ISSUE #249: Include Asserts for field functions into the simulation_base file
E.g.:

https://github.com/MeltPoolDG/MeltPoolDG/blob/6a1a60a28115722d38e86051ceaa742c7b9f6c21/include/meltpooldg/interface/simulation_base.hpp#L300-L304

directly include an assert message:

```cpp
    AssertThrow( ... ,
  ExcMessage("It seems that your SimulationBase object does not contain "        
"a valid initial field function for the level set field. A shared_ptr to your initial field "       
"function, e.g., MyInitializeFunc&lt;dim&gt; must be specified as follows: "))
```

nmuch:
Closes #249

### DIFF
--- a/include/meltpooldg/interface/simulation_base.hpp
+++ b/include/meltpooldg/interface/simulation_base.hpp
@@ -300,31 +300,81 @@ namespace MeltPoolDG
     const std::shared_ptr<Function<dim>> &
     get_initial_condition(const std::string operation_name)
     {
-      return field_conditions_map[operation_name]->initial_field;
+      auto field_conditions = field_conditions_map[operation_name];
+      AssertThrow(
+        field_conditions,
+        ExcMessage(
+          "It seems that your SimulationBase object does not contain a valid initial field function for the \"" +
+          operation_name +
+          "\" field. A shared_ptr to your initial field function, e.g., MyInitializeFunc<dim>; can "
+          "be specified in the SimulationBase::set_field_conditions() function as follows: \n"
+          "this->attach_initial_condition(std::make_shared<MyInitializeFunc<dim>>(...), \"" +
+          operation_name + "\");"));
+      return field_conditions->initial_field;
     }
 
     const std::shared_ptr<Function<dim>> &
     get_source_field(const std::string operation_name)
     {
-      return field_conditions_map[operation_name]->source_field;
+      auto field_conditions = field_conditions_map[operation_name];
+      AssertThrow(
+        field_conditions,
+        ExcMessage(
+          "It seems that your SimulationBase object does not contain a valid source field function for the \"" +
+          operation_name +
+          "\" field. A shared_ptr to your source field function, e.g., MySourceFunc<dim>; can "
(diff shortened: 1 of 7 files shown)


include/meltpooldg/interface/simulation_base.hpp
```diff
@@ -300,31 +300,81 @@ namespace MeltPoolDG
...
+      auto field_conditions = field_conditions_map[operation_name];
+      AssertThrow(
+        field_conditions,
+        ExcMessage(
+          "It seems that your SimulationBase object does not contain a valid source field function for the " +
+          operation_name +
+          " field. A shared_ptr to your source field function, e.g., MySourceFunc<dim>; must "
+          "be specified in the SimulationBase::set_field_conditions() override as follows: \n"
+          "this->attach_source_field(std::make_shared<MySourceFunc<dim>>(...), \"" +
+          operation_name + "\");"));
```
peterrum, line 328:
```suggestion
          "It seems that your SimulationBase object does not contain a valid source field function for the \"" +
          operation_name +
          "\" field. A shared_ptr to your source field function, e.g., MySourceFunc<dim> can "
          "be specified in the SimulationBase::set_field_conditions() function as follows: \n"
          "this->attach_source_field(std::make_shared<MySourceFunc<dim>>(...), \"" +
          operation_name + "\");"));
```

peterrum (approved):
Thanks for tackling this issue. If you want you could also write your own exception as done here: https://github.com/dealii/dealii/blob/13d5d0f88c53b12acd3c0fa9e89419db0413bfc2/include/deal.II/grid/reference_cell.h#L1859-L1860

nmuch:
Thank you, @peterrum, for your review.

> If you want you could also write your own exception

I think that is not worth it because these exceptions are probably only used here and sightly different.

mschreter:
@nmuch thanks a lot for taking over. However, this way the assert would only be thrown if the map is not allocated (and not the specific field function). I've directly made some modifications and removed existing asserts that are obsolete now. I've also introduced the `bool is_optional` argument to the field function, so that the assert is not thrown if the field function is said to be optional. The solution is not perfect but should meet our purposes.

nmuch:
Thank you, @mschreter, for improving this function :+1:  

Have you tried turning on `is_optional` when no field function is provided? What will ` return field_conditions->source_field;` do then?

mschreter:
> Thank you, @mschreter, for improving this function 
> 
> Have you tried turning on `is_optional` when no field function is provided? What will ` return field_conditions->source_field;` do then?

Then we would handle the exception ourselves as we do now in:

https://github.com/nmuch/MeltPoolDG/blob/fb9eb9e77b2f5ad696317225bd8e983aed38008b/source/heat/heat_transfer_problem.cpp#L43-L47

If you have a better solution here, it's very welcome :-).

PR #251, opened 2021-06-09
Generalize evaporative source term in heat equation

mschreter:
This PR generalizes the source term in the heat equation due to evaporation as follows:

```
mDot * h_v * delta
```
mDot is a function of the "old" temperature, however the difference was found to be negligibly small at least for small time step sizes.


source/heat/heat_transfer_operator.cpp
```diff
@@ -507,42 +482,24 @@ namespace MeltPoolDG::Heat
...
-                   *    q_s = m h_v = α_v · h_v · <(T - T_v)> · δ
-                   *                  ^______^                   Γ
-                   *                    evaporation_heat_transfer_coefficient
-                   */
-                  temp = compare_and_apply_mask<SIMDComparison::greater_than_or_equal>(
-                    temp_vals.get_value(q_index),
-                    evapor_data->boiling_temperature,
-                    evaporation_heat_transfer_coefficient * ls_vals.get_gradient(q_index).norm() *
-                      (temp_vals.get_value(q_index) - evapor_data->boiling_temperature),
-                    0.0);
```
peterrum, line 538: I was trying to understand why this else-path can go. If I understood correctly this code path does what `EvaporationModelHardtWondra` + `EvaporationMassFluxOperatorContinuous` does, doesn't it. So removing this code here does not hurt.

peterrum (approved):
Looks good to me! I have one conceptual question, but will nevertheless merge so that you can proceed with refactoring!

source/heat/heat_transfer_operator.cpp
mschreter: Exactly. The difference is that previously we computed the evaporative mass flux from the current `temperature`. Now we consider it as a source term, which is computed using `temperature_old`.

peterrum: Thanks :) Good to know that have not lost the overview yet!

mschreter: :-). The old version would be more exact, but the new version gives as more flexibility :-).

PR #255, opened 2021-06-10
Evaporation line integral: bug fix

mschreter:
This PR:
- computes the distance for the `line_integral` algorithm depending on the reinitialization thickness parameter
- fixes a bug on the `line_integral` algorithm --> the `evaporative_mass_flux` is now zeroed out

This animation shows the film boiling example using the `interface_value` (white level set) and `line integral` (red level set)

animation


source/evaporation/evaporation_mass_flux_operator_interface_value.cpp
```diff
@@ -30,7 +30,11 @@ namespace MeltPoolDG::Evaporation
     , tolerance_normal_vector(
         UtilityFunctions::compute_numerical_zero_of_norm<dim>(scratch_data.get_triangulation(),
                                                               scratch_data.get_mapping()))
-  {}
+  {
+    AssertThrow(scratch_data.get_degree(temp_dof_idx) == scratch_data.get_degree(ls_dof_idx),
+                ExcMessage("This algorithm is currently only supported for the same degree "
+                           "between the heat transfer and the level set."));
```
peterrum, line 36: Could you point out which lines is making problems else?

peterrum (approved):

source/evaporation/evaporation_mass_flux_operator_interface_value.cpp
mschreter: At the moment none.

peterrum: Sounds good. Let's remove it once we have a test!

PR #256, opened 2021-06-10
Move simulation folder

### LINKED ISSUE #164: Move Simulations to top level folder applications

peterrum:
closes #164

FYI @nmuch


CMakeLists.txt
```diff
@@ -90,6 +89,17 @@ if(${DEAL_II_VERSION_MAJOR} GREATER_EQUAL 10)
...
 endif()
 
+# ##############################################################################
+# build meltpooldg library
+# ##############################################################################
+set(libSimulations "meltpooldg-simulations")
+
+file(GLOB_RECURSE sources "${CMAKE_SOURCE_DIR}/simulations/*.cpp")
+
+add_library(${libSimulations} ${sources})
```
mschreter, line 99: So in the `meltpooldg-simulations` all sources are compiled as well? Or do we simply link against the `meltpooldg-lib`?

simulations/simulation_selector.cpp
```diff
@@ -2,31 +2,31 @@
...
-#include <meltpooldg/simulations/spurious_currents/spurious_currents.hpp>
-#include <meltpooldg/simulations/stefans_problem/stefans_problem.hpp>
-#include <meltpooldg/simulations/stefans_problem/stefans_problem1_with_flow_and_heat.hpp>
-#include <meltpooldg/simulations/stefans_problem/stefans_problem2_with_flow_and_heat.hpp>
-#include <meltpooldg/simulations/stefans_problem/stefans_problem_with_flow.hpp>
-#include <meltpooldg/simulations/thermo_capillary_droplet/thermo_capillary_droplet.hpp>
-#include <meltpooldg/simulations/thermo_capillary_two_droplets/thermo_capillary_two_droplets.hpp>
-#include <meltpooldg/simulations/unidirectional_heat_transfer/unidirectional_heat_transfer.hpp>
-#include <meltpooldg/simulations/vortex_bubble/vortex_bubble.hpp>
+#include "advection_diffusion/advection_diffusion.hpp"
```
mschreter, line 7: The path to the folder "simulations" is not necessary anymore?

mschreter (approved):
Thanks Peter, just two minor (mainly CMake-related) questions :-). What about the workflows do they have to adapted as well?

CMakeLists.txt
peterrum: `sources` variables is overridden here. We only compile the files in `simulations` into `libSimulations`. Linking happens at the end.

simulations/simulation_selector.cpp
peterrum: No. These paths are now all relative.

peterrum:
@mschreter I have removed the most of the CMakeList.txt changes because CI had linking errors (in contrast to my local build). I'll take a look at that on an other day. But let's get the destructive movement merged as soon as possible.

mschreter:
Feel free to merge once CI is happy!

Issue #225, opened 2021-06-03
Create a release that is compatible with deal.II 9.3


peterrum:
What about using tags? I think we could create the tags once the following PRs are merged:

- adflo: https://github.com/kronbichler/adaflo/pull/50 -> https://github.com/kronbichler/adaflo/releases/tag/v1.1
- MeltPoolDG: #222/#224
- dealii::VectorTools::point_gradients() is not part of release 9.3. We might need to create a copy from the function there if we intend to tag a version of MeltPoolDG that is working with 9.3.
- The same is true for `dealii::MatrixFreeOperators::CellwiseInverseMassMatrix`

PR #257, opened 2021-06-10
Make compatible with deal.II 9.3 (part 1)

peterrum:
References #225.

At least it compiles. The `CellwiseInverseMassMatrix` operator is still missing. But nevertheless we could merge.

### DIFF
--- a/include/meltpooldg/utilities/legacy.hpp
+++ b/include/meltpooldg/utilities/legacy.hpp
@@ -1,0 +1,120 @@
+/* ---------------------------------------------------------------------
+ *
+ * Author: Magdalena Schreter, TUM, March 2021
+ *
+ * ---------------------------------------------------------------------*/
+
+#pragma once
+
+// a file to ensure compatibility with the 9.3 release of deal.II
+// see also: https://github.com/dealii/dealii/releases/tag/v9.3.0
+#if DEAL_II_VERSION_MAJOR < 10
+
+#  include <deal.II/numerics/vector_tools.h>
+
+namespace dealii ::VectorTools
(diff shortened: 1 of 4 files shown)


source/evaporation/evaporation_mass_flux_operator_thickness_integration.cpp
```diff
@@ -10,6 +10,115 @@
...
+                               ArrayView<const value_type>(evaluation_point_results.data() + ptr[i],
+                                                           n_entries));
+          }
+
+        return unique_evaluation_point_results;
+      }
+  }
+} // namespace dealii::VectorTools
+
+#endif
```
mschreter, line 120: Do you mind to move this function to `vector_tools.hpp`? I would like to keep the physics clean :-).

mschreter (approved):
Thanks Peter, one minor suggestion, but nevertheless feel free to merge.

include/meltpooldg/utilities/legacy.hpp
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Author: Magdalena Schreter, TUM, March 2021
+ *
+ * ---------------------------------------------------------------------*/
+
+#pragma once
+
+// a file to ensure compatibility with the 9.3 release of deal.II
+// see also: https://github.com/dealii/dealii/releases/tag/v9.3.0
+#if DEAL_II_VERSION_MAJOR < 10
```
peterrum, line 11: I went with creating a new file `legacy.hpp`. What do you thinks. This file can be nuked in a week.

PR #259, opened 2021-06-11
Add domain size and initial temperature param

nmuch:
I'm working on a more complex simulation example and stared by generalizing the melt_front_propagation case. I noticed I need to define a several new boundary conditions and decided to create a new case instead of using melt_front_propagation... 

I thought we can use these additional parameters, anyway.

### DIFF
--- a/simulations/melt_front_propagation/melt_front_propagation.hpp
+++ b/simulations/melt_front_propagation/melt_front_propagation.hpp
@@ -35,12 +35,6 @@ namespace MeltPoolDG::Simulation::MeltFrontPropagation
   using namespace dealii;
   using namespace MeltPoolDG::Simulation;
 
-  static constexpr double x_max = 0.6e-3;
-  static constexpr double y_max = 0.2e-3;
-  static constexpr double z_max = 0.2e-3;
-
-  static constexpr double T_0 = 1000.0;
-
   template <int dim>
   class InitialLevelSet : public Function<dim>
   {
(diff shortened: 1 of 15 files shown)


include/meltpooldg/interface/parameters.hpp
```diff
@@ -217,8 +218,10 @@ namespace MeltPoolDG
     bool        do_recoil_pressure             = false;
     number      domain_x_min                   = 0.0;
     number      domain_y_min                   = 0.0;
+    number      domain_z_min                   = 0.0;
     number      domain_x_max                   = 0.0;
     number      domain_y_max                   = 0.0;
+    number      domain_z_max                   = 0.0;
     number      boiling_temperature            = 0.0;
```
mschreter, line 225: These parameters are a result from earlier times, where I also investigated the influence of the domain size. For now it's fine to have them in this section, however we should consider to create in the future a `domain` section, where e.g. a rectangular domain can be chosen and the domain parameters are read.

include/meltpooldg/interface/parameters.hpp
```diff
@@ -161,6 +161,7 @@ namespace MeltPoolDG
     bool                        two_phase                          = false;
     bool                        variable_properties_over_interface = false;
     bool                        solidification                     = false;
+    double                      initial_temperature                = 300.0;
```
mschreter, line 164:
```suggestion
    double                      initial_temperature                = 0.0;
```

mschreter (approved):

mschreter:
Thanks @nmuch, I am closing this PR asap. Then I'll move the `TwoPhaseFlowProblem` to the `MeltPoolProblem`. I hope you are currently not working on bigger changes in the `TwoPhaseFlowProblem`.

include/meltpooldg/interface/parameters.hpp
peterrum: Shouldn't we replace this by bounding boxes?

nmuch:
> @nmuch, I am closing this PR asap. Then I'll move the `TwoPhaseFlowProblem` to the `MeltPoolProblem`. I hope you are currently not working on bigger changes in the `TwoPhaseFlowProblem`.

No I am not working on anything in `TwoPhaseFlowProblem` atm. Go ahead

include/meltpooldg/interface/parameters.hpp
mschreter: I am open to any better suggestions ^^. Could you please elaborate it in more detail what you mean?

peterrum: I am not sure of the purpose of these fields but the info might be extractble by:
```cpp
const auto bb = GridTools::compute_bounding_box(tria);
/*...*/ = bb.get_boundary_points ();
```
But maybe these extends are unrelated to the triangulation?

mschreter: These fields are intended to setup the general domain (`hyper_rectangle`). These parameters define first the triangulation. So unfortunately, I don't think that bounding boxes might help :/.

mschreter:
Let's merge #261 first.

include/meltpooldg/interface/parameters.hpp
peterrum: This means that these parameters are "simulation" specific?

mschreter: exactly.

peterrum: > exactly.

OK. Let's get this merged once the merge conflicts are fixed. However, in the long run case-specific parameters should be handled differently. In hyper.deal, parameter files are read multiple times and only the relevant parameters are extracted when needed. This allows to have case-specific parameters: https://github.com/hyperdeal/hyperdeal/blob/31a3fc2e2b3b948daa0ecce9e2c34c0a51cc43de/examples/vlasov_poisson/cases/torus_hyperball.h#L88-L100

nmuch:
@mschreter I rebased this PR on the current master and applied the concept from #272.

mschreter:
> @mschreter I rebased this PR on the current master and applied the concept from #272.

Thanks a lot for your efforts @nmuch, looks very good and clean now! Sorry for the inconveniences that you have to rebase multiple times :-/. Next time we'll try to organize the merging to be more straightforward and less conflicting :-). At least we have now a clear track how to distinguish between simulation-specific and general parameters right from the start :-). I am merging once CI is happy.

PR #266, opened 2021-06-11
Make compatible with deal.II 9.3 (part 2)

peterrum:
References #225.


mschreter:
Thanks Peter, so you have an explanation of the linking error in build adaflo?

peterrum:
> so you have an explanation of the linking error in build adaflo?

Unfortunately, no. Since it is unrelated to this PR (the tests of other PRs are also failing), I would suggest to merge. I'll take a look at the issue later on.

PR #272, opened 2021-06-15
Add simulation specific parameters

mschreter:
This PR introduces a particular function to declare simulation specific parameters `domain_x_min` (see `recoil_pressure.hpp`). I did not want to read the base parameters twice (note: child function cannot be executed from the base class constructor) to avoid potential conflicting behavior, so I declared a separate `ParameterHandler` for the user-specific parameters.

@nmuch: Could you please adapt #259 to this concept?

### DIFF
--- a/source/interface/parameters.cpp
+++ b/source/interface/parameters.cpp
@@ -6,7 +6,8 @@ namespace MeltPoolDG
   void
   Parameters<number>::process_parameters_file(const std::string &parameter_filename)
   {
-    add_parameters();
+    AssertThrow(!parameters_read, ExcMessage("The parameters are already read once."))
+      add_parameters();
 
     check_for_file(parameter_filename);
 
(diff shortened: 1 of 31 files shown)


source/interface/parameters.cpp
```diff
@@ -51,13 +52,37 @@ namespace MeltPoolDG
...
+          material.two_phase_properties_transition_type =
+            MaterialData<number>::TwoPhasePropertiesTransitionType::smooth;
+        else if (flow.variable_properties_over_interface == "consistent_with_evaporation")
+          material.two_phase_properties_transition_type =
+            MaterialData<number>::TwoPhasePropertiesTransitionType::evaporation;
+        else
+          AssertThrow(false,
+                      ExcMessage("Unknown variable properties over interface type \"" +
+                                 flow.variable_properties_over_interface + "\"! Abort..."));
+
```
mschreter, line 85: @nmuch: I had to move this section from `add_parameters()` to `process_parameter_file()` since at the stage of `add_parameters()` the `json`-file is not parsed yet. The parameter parsing is done here: https://github.com/MeltPoolDG/MeltPoolDG/blob/9ceea6d8fbb1434acf8b67f52bba3d6b0138ebc5/source/interface/parameters.cpp#L16-L21 The assignment only worked by chance, since up to now we read the parameters twice.

peterrum (approved):
Thanks :1st_place_medal:

PR #273, opened 2021-06-15
Add singularity scripts

mschreter:


scripts/singularity/meltpooldg_dealii_docker.def
```diff
@@ -0,0 +1,83 @@
...
+  fi
+  #--------------------------------------
+  # setup system
+  #--------------------------------------
+  set -ex
+  apt update
+
+  apt-get -y install apt-utils apt-file less make wget openssh-client
+  apt-get -y install openmpi-bin  infiniband-diags ibverbs-utils libibverbs-dev
+  apt-get -y install libipathverbs1 libmlx4-1 libmlx5-1 libmthca1
```
peterrum, line 49: Is this all needed? Isn't that already loaded in `dealii/dealii:master-focal`?

peterrum (approved):

scripts/singularity/meltpooldg_dealii_docker.def
mschreter: That's a good point. I'll eliminate some lines.

mschreter: I've removed a couple of lines.

peterrum (approved):

PR #274, opened 2021-06-15
Introduce multiplicity in line integral

mschreter:
This PR:

- computes multiplicity to evaluate the evaporative mass flux via thickness integration
- adds the check if a point is outside the bounding box within `generate_points_along_normal_vector`.

In the following picture we see the resulting evaporative mass flux. In this example periodic BC are applied between the right and the left face. This would mean that the values should be the same on the two faces, however they are not (on the right face they are exactly zero).

image

### DIFF
--- a/include/meltpooldg/utilities/utility_functions.hpp
+++ b/include/meltpooldg/utilities/utility_functions.hpp
@@ -483,6 +483,10 @@ namespace MeltPoolDG
 
       global_points_normal_to_interface_pointer = {0};
 
+      const auto bounding_box = GridTools::compute_bounding_box(dof_handler_ls.get_triangulation());
+      const auto boundary_points = bounding_box.get_boundary_points();
+
+
       const auto collect_points_along_normal = [&](const auto &                 cell,
                                                    const auto &                 real_points,
                                                    const auto &                 unit_points,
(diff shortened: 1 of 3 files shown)


peterrum:
> (on the right face they are exactly zero).

What about `AffineConstraints::distribute()`? `AffineConstraints::distribute_local_to_global()` never writes into constrained values with the consequence that these values are zero. Alternatively, you can use an unconstrained `AffineConstraints` object!

include/meltpooldg/utilities/utility_functions.hpp
```diff
@@ -548,7 +552,22 @@ namespace MeltPoolDG
                                               max_distance_per_side,
                                               n_inc_per_side,
                                               bidirectional);
-
+            // check if point is outside domain and if so then project it back to
+            // the domain
+            //
+            // @todo: compute intersection between normal and bounding box and use
+            // this point
```
peterrum, line 559: Indeed the following function looks a bit odd. Why not simply ignore the points outside? Alternatively, for intersection one could use boost geometry.

mschreter:
> > (on the right face they are exactly zero).
> 
> What about `AffineConstraints::distribute()`? `AffineConstraints::distribute_local_to_global()` never writes into constrained values with the consequence that these values are zero. Alternatively, you can use an unconstrained `AffineConstraints` object!

Thanks a lot, that was the issue. I assume there will be some remaining similar issues in case of PBC :-/.

source/evaporation/evaporation_mass_flux_operator_thickness_integration.cpp
```diff
@@ -229,6 +237,22 @@ namespace MeltPoolDG::Evaporation
         remote_point_evaluation.template process_and_evaluate<double>(mass_flux_val,
                                                                       buffer,
                                                                       broadcast_function);
+
+        // take care dofs shared by multiple geometric entities
+        evaporative_mass_flux.compress(VectorOperation::add);
+        vector_multiplicity.compress(VectorOperation::add);
+
+        vector_multiplicity.update_ghost_values();
+        evaporative_mass_flux.update_ghost_values();
```
peterrum, line 246:
```suggestion
```

peterrum:
> Thanks a lot, that was the issue. I assume there will be some remaining similar issues in case of PBC :-/.

Hope not. This is more a visualization issue. If you use `AffineConstraints` consistently, it should handle PBCs for you as expected!

peterrum (approved):

peterrum:
> In the following picture we see the resulting evaporative mass flux.

What about updating the picture :P

include/meltpooldg/utilities/utility_functions.hpp
mschreter: I am erasing the points now when they are outside.

mschreter:
> > In the following picture we see the resulting evaporative mass flux.
> 
> What about updating the picture :P

There you go :-)

image

peterrum:
Thanks :)

PR #276, opened 2021-06-17
Enable time-dependent update of zero velocity constraints

mschreter:
This PR enables 

- `heat_transfer` BC in the recoil pressure example.
- the time-dependent update of zero velocity constraints in the solid phase. This can be enabled by setting 
```json
    "mp set velocity to zero in solid": "true"
```
in the parameter file. The results looks now as follows for the coupled thermo-two-phase-flow melt pool problem (blue line marks the liquid domain):

@nmuch FYI

image

### DIFF
--- a/simulations/recoil_pressure/recoil_pressure.hpp
+++ b/simulations/recoil_pressure/recoil_pressure.hpp
@@ -22,6 +22,8 @@ namespace MeltPoolDG
   {
     namespace RecoilPressure
     {
+      const double T_initial = 500.;
+
       using namespace dealii;
 
       template <int dim>
@@ -157,12 +159,17 @@ namespace MeltPoolDG
         void
         set_boundary_conditions() override
         {
+          types::boundary_id lower_bc = 0;
+          types::boundary_id upper_bc = 0;
+          types::boundary_id left_bc  = 0;
+          types::boundary_id right_bc = 0;
+
           if (this->parameters.mp.do_evaporation)
             {
-              const types::boundary_id lower_bc = 1;
-              const types::boundary_id upper_bc = 2;
-              const types::boundary_id left_bc  = 3;
-              const types::boundary_id right_bc = 4;
+              lower_bc = 1;
+              upper_bc = 2;
+              left_bc  = 3;
+              right_bc = 4;
 
               if constexpr (dim == 2)
                 {
@@ -190,35 +197,11 @@ namespace MeltPoolDG
 
               this->attach_dirichlet_boundary_condition(
                 upper_bc, std::make_shared<Functions::ConstantFunction<dim>>(-1.0), "level_set");
-              if (this->parameters.laser.heat_source_model != "Analytical")
-                {
-                  this->attach_dirichlet_boundary_condition(
-                    lower_bc,
-                    std::make_shared<Functions::ConstantFunction<dim>>(500),
-                    "heat_transfer");
-                  this->attach_dirichlet_boundary_condition(
-                    upper_bc,
-                    std::make_shared<Functions::ConstantFunction<dim>>(500),
-                    "heat_transfer");
-                  this->attach_dirichlet_boundary_condition(
-                    left_bc,
-                    std::make_shared<Functions::ConstantFunction<dim>>(500),
-                    "heat_transfer");
-                  this->attach_dirichlet_boundary_condition(
-                    right_bc,
-                    std::make_shared<Functions::ConstantFunction<dim>>(500),
-                    "heat_transfer");
-                }
             }
           else if (this->parameters.base.problem_name == "melt_pool")
             {
               this->attach_no_slip_boundary_condition(0, "navier_stokes_u");
               this->attach_fix_pressure_constant_condition(0, "navier_stokes_p");
-              if (this->parameters.laser.heat_source_model != "Analytical")
-                {
-                  this->attach_dirichlet_boundary_condition(
-                    0, std::make_shared<Functions::ConstantFunction<dim>>(500), "heat_transfer");
-                }
             }
           else
             AssertThrow(false, ExcNotImplemented());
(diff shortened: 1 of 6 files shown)


simulations/recoil_pressure/recoil_pressure.hpp
```diff
@@ -190,14 +190,42 @@ namespace MeltPoolDG
 
               this->attach_dirichlet_boundary_condition(
                 upper_bc, std::make_shared<Functions::ConstantFunction<dim>>(-1.0), "level_set");
+              if (this->parameters.laser.heat_source_model != "Analytical")
```
peterrum, line 193: What is the logic here:
```cpp
if (this->parameters.laser.heat_source_model != "Analytical"){}
  else if (this->parameters.base.problem_name == "melt_pool")
  {
    if (this->parameters.laser.heat_source_model != "Analytical"){/*dead code!?*/}
  }
```

source/melt_pool/melt_pool_operation.cpp
```diff
@@ -421,6 +437,8 @@ namespace MeltPoolDG::MeltPool
 
     UtilityFunctions::check_constraints(level_set_dof_handler, level_set_constraints);
 
+
+
```
peterrum, line 441:
```suggestion
```

mschreter:
Using this setup I was able to simulate the following laser melting example. It is weird that at the tip of the melt splash the material is already cooled down again below the melting temperature such that a "solid domain" forms. Apart from that I am quite happy with these very first results :-).

animation

source/melt_pool/melt_pool_problem.cpp
```diff
@@ -448,6 +454,7 @@ namespace MeltPoolDG::Flow
        */
 #ifdef MELT_POOL_DG_WITH_ADAFLO
     dynamic_cast<AdafloWrapper<dim> *>(flow_operation.get())->reinit_1();
+    flow_velocity_constraints_no_solid.copy_from(flow_operation->get_constraints_velocity());
```
peterrum, line 457: I am still a bit puzzled by this line. Is `flow_velocity_constraints_no_solid` really necessary, I guess so. But why did it work (or did I have the wrong impression)?

source/melt_pool/melt_pool_problem.cpp
```diff
@@ -25,7 +25,9 @@ namespace MeltPoolDG::Flow
         //
         // @todo: alternative (better performing) solution?
         if (base_in->parameters.mp.set_velocity_to_zero_in_solid)
-          scratch_data->build();
+          {
+            scratch_data->build();
+          }
```
peterrum, line 30:
```suggestion
          scratch_data->build();
```
;)

source/melt_pool/melt_pool_problem.cpp
mschreter: `flow_velocity_constraints_no_solid` stores the constraints resulting from "real" boundary conditions (dirichlet, no-slip etc.). In 
https://github.com/MeltPoolDG/MeltPoolDG/blob/292abc2197ba337b13e4ced911e1a5717f22a6a4/source/melt_pool/melt_pool_operation.cpp#L315-L374

we merged the flow velocity constraints with the solid constraints **every time step**. This means that if a point was rigid (velocity set to zero = solid) once in history it will never be deformable (velocity, liquid or gas) again. So instead of overwriting the entire flow velocity constraints everytime, we use the "real" constraints stored in `flow_velocity_constraints_no_solid` and then merge potential solid constraints into it. Up to now it has worked for our examples, since we had there a never changing solid domain.

source/melt_pool/melt_pool_problem.cpp
mschreter: Aah you got me, this PR was a rather quick-and-dirty one, which obviously not passes the borders to hungary ^^. I'll put some effort into cleaning it up later.

source/melt_pool/melt_pool_problem.cpp
peterrum: This was what I thought. Thanks for the explanation! I guess if you would have enabled AMR in each time step the results would be the same, since the constrains are nuked incl. the old solid dofs.

simulations/recoil_pressure/recoil_pressure.hpp
```diff
@@ -190,14 +190,42 @@ namespace MeltPoolDG
...
+                {
+                  this->attach_dirichlet_boundary_condition(
+                    lower_bc,
+                    std::make_shared<Functions::ConstantFunction<dim>>(500),
```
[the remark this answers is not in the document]
mschreter: initial temperature field.

simulations/recoil_pressure/recoil_pressure.hpp
mschreter: I've improved it.

mschreter:
I've included the changes.

simulations/recoil_pressure/recoil_pressure.hpp
```diff
@@ -190,43 +197,37 @@ namespace MeltPoolDG
...
             }
           else if (this->parameters.base.problem_name == "melt_pool")
             {
               this->attach_no_slip_boundary_condition(0, "navier_stokes_u");
               this->attach_fix_pressure_constant_condition(0, "navier_stokes_p");
-              if (this->parameters.laser.heat_source_model != "Analytical")
-                {
-                  this->attach_dirichlet_boundary_condition(
-                    0, std::make_shared<Functions::ConstantFunction<dim>>(500), "heat_transfer");
-                }
```
peterrum: Sorry for the noise.  I think I prefer how it used to be (minus the "500" constants). I didn't see that in one case you have 4 bids and in the other case only a single one.

peterrum (approved):
Thanks a lot!

simulations/recoil_pressure/recoil_pressure.hpp
mschreter: No problem. I've restored it the old way :-).

peterrum (approved):

PR #277, opened 2021-06-20
Simplify CI

peterrum:
by
- mounting directly the Docker image
- introducing a strategy matrix
- integrating the nightly tests


mschreter:
Thanks!

